PR #25, opened 2018-12-05
Initial concentrations for species 

zoltuz:
initial concentration is written to the sbml file, if the corresponding species' initial concentration was not set. A simple example and Unit test were added.

### DIFF
--- a/biocrnpyler/chemicalreactionnetwork.py
+++ b/biocrnpyler/chemicalreactionnetwork.py
@@ -11,9 +11,10 @@ class Specie(object):
 
     """
 
-    def __init__(self, name, type="", attributes=[]):
+    def __init__(self, name, type="", attributes=[], initial_concentration=None):
         self.name = name
         self.type = type
+        self.initial_concentration = initial_concentration
 
         if attributes is None:
             attributes = []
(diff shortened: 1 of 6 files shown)


ayush9pandey:
One bit of information that we should keep in mind is that there are two ways in which amounts can be set to species in SBML. One is `setInitialConcentration` and the other is `setInitialAmount`. The difference between the two is dictated by the attribute that each species has called `hasOnlySubstanceUnits`. I am copying the text from SBML documentation here -

>simply setting 'initialAmount' or 'initialConcentration' alone does not determine whether a species identifier represents an amount or a concentration when it appears elsewhere in an SBML model. The role of the attribute 'hasOnlySubstanceUnits' is to indicate whether the units of the species, when the species identifier appears in mathematical formulas, are intended to be concentration or amount. The attribute takes on a boolean value. In SBML Level 3, the attribute has no default value and must always be set in a model; in SBML Level 2, it has a default value of False.
 

TL;DR - Set `hasOnlySubstanceUnits` attribute in Species to False and use `setInitialConcentration` to give concentrations to species. Otherwise, set that attribute to True. 

All of this is important because biocrnpyler is generating SBML models and these SBML models could be used in any other SBML compatible tool (say for simulation of this SBML model), most simulators look into the units of species and parameters and do their calculations accordingly, so we need to ensure that we are using these attributes correctly. (Note : bioscrape does not care about units, so simulations in bioscrape would not be affected).

PR #41, opened 2019-08-01
adding unit tests for all top level classes, updating gitingore, setting up docs folder for Sphinx

zoltuz:
I've updated the git ignore and added a config file for  Sphinx documentation generator. 
(Also one file was updated with minor style changes)


zoltuz:
I've also added a unittest on the component class and some minor changes in the component class . Plus some changes in the travisCI config.

zoltuz:
I've added unit tests for all the top level classes*, all of the are passing at the moment:
https://travis-ci.org/zoltuz/biocrnpyler/builds 
An example run can be view here:
https://travis-ci.org/zoltuz/biocrnpyler/jobs/568308266

*except: ComplexSpecies, we need to talk about this class at some point.

murrayrm:
You should also add the webhooks required so that the unit tests get run across all new PRs.

ayush9pandey:
We can merge this pull request and I can fix the two test errors with SBML in my pull request. 

Also, just as a note, on running Python [Coverage ](https://coverage.readthedocs.io/en/v4.5.x/) package, it is reporting that our current unit test framework covers around 50% of the code - a good start in my opinion, and we can work on improving it later.

zoltuz:
@ayush9pandey thanks for checking the PR! The two SBML errors are a bit strange, I guess something has change in the underlying code. 
The third error is python 3.6 specific, somehow the way I test for file content is not working in python 3.6, but works well in python 3.7... (I will look into it later)

ayush9pandey:
Yeah, I will fix the two SBML errors with my PR.

PR #68, opened 2020-05-04
Combinatorial Promoter

dr3y:
small bug fix: CombinatorialPromoter was not setting self.tx_capable_list if you gave it a good looking list


dr3y:
I added a CombinatorialPromoter ipynb

zoltuz:
Hi there, 
The checks in TravisCI failed because of this:
File "/home/travis/build/BuildACell/BioCRNPyler/biocrnpyler/dna_construct.py", line 9, in <module>
    from matplotlib import cm
ModuleNotFoundError: No module named 'matplotlib'

You need to add matplotlib to the setup.py as a dependency, so it gets installed when a user tries to use biocrnpyler.

PR #73, opened 2020-05-13
Combinatorial promoter

dr3y:
I cleaned up this branch a bit so the only added stuff should be pertaining to combinatorial promoter and combinatorial promoter examples.


biocrnpyler/mechanism.py
```diff
@@ -331,11 +332,111 @@ def update_reactions(self, s1, s2, component = None, kb = None, ku = None, part_
             ku = component.get_parameter("ku", part_id = part_id, mechanism = self)
         if component == None and (kb == None or ku == None):
             raise ValueError("Must pass in a Component or values for kb, ku.")
-
-        complex = ComplexSpecies([s1, s2])
+        if(complex==None):
```
zoltuz, line 335: Hi there,
I'd suggest that to test complex object points to None use this instead: "complex is None", because if someone implements the equality operator on the Complex class then this code will cause trouble. 
This is true for other cases where you test for None.
More info here:
https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

biocrnpyler/mechanism.py
```diff
@@ -331,11 +332,111 @@ def update_reactions(self, s1, s2, component = None, kb = None, ku = None, part_
...
+        else:
+            myspecies = ComplexSpecies(complexed_species_list)
+        return myspecies
+    def update_species(self,binders,bindee,cooperativity=None,\
+                            component = None, part_id = None, **kwords):
+        cooperativity_dict = {}
+        for binder in binders:
+            binder_partid = part_id+"_"+binder.name
+            if ((cooperativity == None) or (type(cooperativity)==dict and binder_partid not in cooperativity) \
+                                                                                        and (component != None)):
```
zoltuz, line 369: some here as well: component is not None

dr3y: ok I will fix it in my code but it seems like this is present in other mechanisms too

zoltuz: I see I'll try to weed them out from the other places.

PR #79, opened 2020-05-26
refactoring tests: adding documentation, optimizing package import and adding regexp check for errors and warnings

zoltuz:
this update
- refactors the tests: adds some documentation, removes multiple package imports and adds regexp check for error messages
- adds type hints to some of the functions
- replace the equivalence test for None (from a == None  to A is None)
- removing *.orig files that got into the repo by mistake

Issue #76, opened 2020-05-15
Parameter files are somehow "global"


WilliamIX:
@Zoltan this one is for you because it has to do with some magic you worked in the Parameter class that I don't understand - I've hacked it, but please help find a better solution (I know there is one).

Issue: If I load one parameter file for a given Mixture, but then I load a different parameter file for a different mixture, the parameters loaded in the first Mixture are still around. This means I cannot have separate parameter files for separate models that I run in the same Jupyter notebook without running into conflicts which is a major issue. I think the issue is that Parameter is a "global instance" of a class we use over and over again - but I'm not 100% sure this is the case.

I have "fixed" this in a hacky way by overwriting the parameter file as an empty dictionary {} whenever Parameter.create_parameter_dictionary is called by default. The old behavior can be restored by setting overwrite = False in this method. I am not sure whether this hacky "fix" still allows Components to overwrite subsets of their Mixture's parameters or not.

A larger question is why do we store parameters inside the instance of a class? Originally, these functions were meant to be helper functions to generate dictionaries which store parameters inside of Components and Mixtures - the goal being that different Components and Mixtures can have custom sets of parameters (perhaps defined by different parameter files or user assigned dictionaries). If all the parameters go to a single dictionary, these kinds of custom parameters will no longer work the same way.

zoltuz:
Have a look at #87, this 'global' behavior also affects all other objects where there's a mutable default argument, e.g.:
```python
class Component(object):

    def __init__(self, name: Union[str, Species],
                 mechanisms={},  # custom mechanisms gets initialized ONCE per run, not one per object initialization!
```

zoltuz:
Based on the description of the issue, #87 and #117 took care of the problem. Reopen if not.

PR #87, opened 2020-06-02
Bug fix parameter loading

zoltuz:
This bug fix fixes the problem described in #76, namely parameter dictionary is initialized one per run and all objects share the same parameter dictionary, due to default mutableable argument behavior.

- tests are adjusted to the new code logic


zoltuz:
@WilliamIX we need to close #79 before this gets accepted, otherwise it might mess up the master branch. Sorry for the inconvenience.

PR #89, opened 2020-06-04
Matthieu Pull Request MTX and MTL mechanisms

mkratz:


biocrnpyler/mechanism.py
```diff
@@ -605,6 +605,136 @@ def update_reactions(self, dna, component = None, kexpress = None,
             protein = Species(dna.name, material_type="protein")
         rxns = [Reaction(inputs=[dna], outputs=[dna, protein], k = kexpress)]
         return rxns
+    
+class multi_tx(Mechanism):
```
zoltuz, line 609: Hi Matthieu,
Would you mind adding at least a one-line comment to this and other classes in your PR, so users can know how to use/what to expect from this class? Many thanks!

mkratz: Hey Zoltuz,
Should I just add the description to the doc-string of the class or somewhere else? I can also refer them to the demo notebook in this PR.

zoltuz: docstring of a class would be the best place. Thanks!

mkratz: Just updated with docstring and directing to demo nb for detailed mechanism and demonstration.

WilliamIX:
This looks really good @mkratz. Could you please make the following two changes before I approve the merge:
1. Update your Mechanisms so that the occupancy is a parameter they search for (as opposed to an argument of the update_species and update_reactions function)? This will allow different occupancies for different species (using part_id).
2. Instead of creating new species to represent bound complexes, such as: 
     Species(self.pol.name + 'x' + dna.name + '_closed' + '_' + str(n), material_type='dna')
   please use the ComplexSpecies([species 1, species2...]) class.

mkratz:
> This looks really good @mkratz. Could you please make the following two changes before I approve the merge:
> 
> 1. Update your Mechanisms so that the occupancy is a parameter they search for (as opposed to an argument of the update_species and update_reactions function)? This will allow different occupancies for different species (using part_id).
> 2. Instead of creating new species to represent bound complexes, such as:
>    Species(self.pol.name + 'x' + dna.name + '_closed' + '_' + str(n), material_type='dna')
>    please use the ComplexSpecies([species 1, species2...], name = None) class. By default name will
>    be something similar to what you wrote above but without the "closed". You can also use custom
>    names if you prefer.

Hey Will, I've made change 2. For change 1, I have the issue that I need max_occ for the update_species, as it defines the number of complex species i have. From looking at promoter, it doesn't pass component or part id. As such,  I don't think I could load max_occ as a parameter in update_species, I would probably have to manually load it, or create a new promoter component that passes the component and parameter id in update_species....

WilliamIX:
Good point @mkratz - I think what you are really suggesting is we should make update_species(....) also take the Component and part_id as a parameter in case we want to algorithmically produce species based upon parameter values which seems reasonable. I will add this as an issue and hopefully make the change next week across the board.

mkratz:
@WilliamIX Yeah, that seems like the best option. I'll push the complex species change along with the max_occ change once you implement that.

WilliamIX:
@mkratz  just a heads up that I think I made the changes you need to make occupancy a parameter. If you update to the newest version of master you should get these options - please let me know if everything works. It would be awesome if you could also move your new mechanisms into the mechanisms_txtl.py file (mechanism.py now only houses the base Mechanism class) and verify that everything works for you. Thanks!

WilliamIX:
@mkratz there apparently seem to be conflicts - could you try to merge master into your branch then redo this request?

mkratz:
@WilliamIX Should be good now

zoltuz:
Unfortunately, the build failed because of some syntax error caused by the merge, e.g.

E     File "/home/travis/build/BuildACell/BioCRNPyler/biocrnpyler/mechanisms_txtl.py", line 295
E       <<<<<<< HEAD
E        ^
E   SyntaxError: invalid syntax

zoltuz:
Almost there! Your example notebook doesn't pass yet:
Screenshot 2020-06-15 at 23 35 02

You need to make the bioscrape simulations conditionals, just like in the other examples.

mkratz:
@zoltuz thanks for the help! Hopefully it passes now... It's weird as I've never had to import bioscrape when running this locally, even now

zoltuz:
The root of the problems is that you have bioscrape installed on your local computer, but BioCRNPyler needs to build and pass all the tests without bioscrape. To do so, you need to make all bioscrape calls conditional in your example notebook. 
Sorry for the trouble!

zoltuz:
The best way to test this is the following:
1, create a virtual environment https://realpython.com/lessons/creating-virtual-environment/
2, activate your virtual env
3, checkout your github code with git clone
4, run python setup.py test 
if you get no error, then it works (unless you installed bioscrape on the system level)

mkratz:
@zoltuz Eureka! it built!

PR #90, opened 2020-06-06
new mechanism type

zoltuz:
Hi guys,

I wanted to pick your brain on an idea: currently, mechanism_type is a string and all the checking and looking up is string-based, e.g. 
```python
'rna_degradation' in self.mechanisms
```

An alternative would be a class type-based system: Each specialised Mechanism has two parents:
```python
class SimpleTranscription(Mechanism, Transcription):
   pass
```
This sets up a system where I can check these:
```python
sm = SimpleTranscription()
instanceof(sm,Mechanism)
instanceof(sm,Transcription)
``` 
using some magic under the hood, you can do this:
```python
if RNADegradation in self.mechanisms
            deg_mech = self.mechanisms[RNADegradation]
```
Two important things are happening here:
 - the list of mechanisms (which is a special subclass of list) can be check if it contains a type called RNADegradation
 - The actual instance of a mechanism can be looked up by the type RNAdegradation.

When you define the mechanism of a mixture, you can list them:
```python
mech_tx = Transcription_MM(rnap = self.rnap.get_species())
mech_tl = Translation_MM(ribosome = self.ribosome.get_species())
mech_rna_deg = Degredation_mRNA_MM(nuclease = self.rnaase.get_species())
default_mechanisms = MechanismList((mech_tx, mech_tl, mech_rna_deg))
```
(This MechanismList() can be moved deeper in the code, so this level the user just defines a list of mechanisms)

More importantly, the rest of the code remains the same:
 ```python
def update_mechanisms(self, mixture_mechanisms={}, mechanisms={},
                          overwrite_custom_mechanisms=True):

        for mech_type in mixture_mechanisms:
            self.mechanisms[mech_type] = mixture_mechanisms[mech_type]
```
this above works the same as before (just under the hood it looks up types and not string matches)

This PR is just a skeleton, definitely do not merge it.

Let me know what you do think!

### CODE REVIEW

**Comment by WilliamIX** [comment]
Yes, I think this is a great idea! I'm about to update the master branch with a refactoring of the file structure and then add a variety of classes from which should actually make this easier to flesh out.

A couple of thoughts:
Would it be possible to use this to enforce the notion of a reaction schema somehow? What I'm imagining is kind of like an Interface in Java if you are familiar with that. For example, a Transcription Mechanisms would require that the update_species and update_reaction have a particular minimum set of keywords (like gene and transcript) but still allow for additional keywords (like polymerase) which could be ignored if they are passed into a mechanism that doesn't need them?

**Comment by zoltuz** [comment]
William's example to consider:
Context (Transcription) + Enzymatic Reaction
G + RNAP <--> G:RNAP --> G + RNAP + T
What if, I only want Enzymatic reaction, but not the context with it
 E + S <--> E:S --> E + S + P 

One solution could be:
```python
class MM_Cat(MM_Copy, Catalysis):
     pass
```

PR #94, opened 2020-06-15
test jupyter nb examples

zoltuz:
Hi all,
I've added tests for Jupyter Notebook via a pytest plugin. That means that the unit test backend has been replaced, but all the current code for the unit tests works fine.  
The only issue is with the combinatorial promoter example right now:
Screenshot 2020-06-15 at 19 46 30
It looks like it tries to use pandas, seaborn and bioscrape (although bioscrape appears in grey, which means that it not actually called in the file)
Screenshot 2020-06-15 at 19 49 31
@dr3y Could you look into this? The other Jupyter examples use a try-catch to exclude bioscrape
Currently pandas and seaborn are not part of the dependency list. We could add them, but a long list of dependencies may hurt us in the long run. 

Once the problem with Combinatorial promoter example gets resolved, this PR can be merged.


dr3y:
I am using sns.heatmap and pd.pivot_table to produce the heatmap

zoltuz:
That's fine. The question is shall we make pandas and seaborn an official dependency of bioCRNplyer? Alternatively, you can put the code, which calls these packages into a try-catch (just like in the case of bioscrape).

dr3y:
I can do the try catch, then if you don't have the package the notebook doesn't run at all basically. Is that ok?

zoltuz:
Fine with me for now. @WilliamIX @ayush9pandey any opinion on this?

dr3y:
I put a pull request to your branch

zoltuz:
Thanks!

WilliamIX:
I don't think we should formally require more plot libraries for BioCRNpyler at this point. I think either having a single example for testing purposes and more examples in try / catch statements is OK. I also think it would be possible to switch all of the heatmaps in Andres' examples with Matplotlib heatmaps.

zoltuz:
@WilliamIX all good now, thanks to Andrey! You can merge the code if you want. It will almost double the code coverage thanks to the checks on the Jupyter notebooks.

WilliamIX:
This is very cool and helpful! I will merge now. 

Is there a way to see which specific classes and methods aren't being called? This will help me build more examples and tests.

zoltuz:
Yes, there is:
if you go on codecov.io using the link in the codecov report above, you can select individual files and see what lines are covered by the tests (green) and what not (red). For example, in components:
https://codecov.io/gh/BuildACell/BioCRNPyler/src/master/biocrnpyler/component.py

PR #102, opened 2020-06-20
DNA_construct

dr3y:
generic DNA construct simulation which predicts RNA and protein production based on a sequence of DNA parts.

Check out DNA construct examples .ipynb!


biocrnpyler/dna_part.py
```diff
@@ -0,0 +1,70 @@
+################################################################
```
zoltuz, line 1: There's a standard header on the top of the new files with licence info, please add it.

zoltuz:
Hi Andrey, 
I think these three commits could be merged together:
Screenshot 2020-06-21 at 10 17 53
To do so:
 1. checkout your branch
 2. run 
```
git rebase -i HEAD^^^
```
Then you see the screenshot above (depending on your default editor in terminal)
To merge the commits, you can use _squash_ or _fixup_ 
Screenshot 2020-06-21 at 10 19 18
hopefully, it can merge without merge conflict, if not you need to handle the conflicts. For each conflict, after resolving it, you need to run
```
git add file_name
# and 
git rebase --continue 
```
If you get lost
```
git rebase --abort
```
to start over.

Because these steps rewrites history, you need to force push your changes to GitHub:
```
git push -f
```

zoltuz:
One more thing, there's one error, which is not present in the dev branch, but most probably this PR introduces: 
Screenshot 2020-06-21 at 12 55 36
This is generated in ' examples/Plotting Examples.ipynb::Cell 2'

You can reproduce this after running
```
python setup.py test
```
on your branch

zoltuz:
Hi Andrey,

 please rebase your branch to dev, so you get the latest updates: 
```
git fetch upstream/dev
git checkout DNA_construct_final
git rebase upstream/dev
```
Once you have that you need to force push:
```
git push -f
```
After these steps, Travis should be able to run properly.

biocrnpyler/dna_part.py
dr3y: where can i find this header?

zoltuz: Here you go:
# Copyright (c) 2020, Build-A-Cell. All rights reserved.
# See LICENSE file in the project root directory for details.

dr3y:
ok i am doing this rebase thing and it has too many conflicts. i don't understand why squashing those commits would create conflicts. are you sure this is worth it?

zoltuz:
> ok i am doing this rebase thing and it has too many conflicts. i don't understand why squashing those commits would create conflicts. are you sure this is worth it?

Probably doesn't worth it, we can do a squash and merge at the end:
Screenshot 2020-06-28 at 22 19 34

WilliamIX:
**Some thoughts on how to deal with binding location on DNA constructs.**

Issue description: DNA constructs (and other future objects) will require that a species bind to a specific location on the piece of DNA. This means that BoundDNA = ComplexSpecies([DNA, Binder]) will not work because the Binder could bind to multiple locations on the DNA. This issue has currently been solved with the make_complex function which does some fancy logic to bind things to the appropriate place on OrderedComplexSpecies. However, this solution is not necessarily generalizable to the case where there are many possible binding locations and different mechanisms require different locations along the same species.

Suggested Solution: Introduce a new Species class SubSpecies(Species). SubSpecies can ONLY exist inside a ScaffoldSpecies (there is probably a better name) which is effectively an OrderedComplexSpecies made up of SubSpecies. Importantly, SubSpecies know about their containing ScaffoldSpecies and use it's name. However, when a SubSpecies is passed into ComplexSpecies or OrderedComplexSpecies, the resulting Species object is automatically placed inside of the ScaffoldSpecies which contains the Subspecies. 

I will illustrate this with an example of two binding sites B1 and B2 in a polymer P which can be bound to A1 and A2, respectively. Note that this is a combinatorial system with 4 possible states.

`B1 = SubSpecies("B1")'
`B2 = SubSpecies("B2")` 
`P = ScaffoldSpecies ([B1, B2])` ==> B1 and B2 can now only be used in a reaction or ComplexSpecies after they are placed inside a Parent Species. If B1 or B2 were used before this line of code, an error would occur.
`C1 = ComplexSpecies([A1, B1])` ==> `Returns ScaffoldSpecies([ComplexSpecies([A1, B1]), B2])` Notice that A1 binds to B1 inside of the ScaffoldSpecies that contains B1 and B2.
`C2 = ComplexSpecies([A2, B2])` ==> `Returns ScaffoldSpecies ([B1, ComplexSpecies([A2, B2])])`
`C3 = ComplexSpecies([A1, C1.species[B1])` ==> `Returns ScaffoldSpecies([ComplexSpecies([A1, B1]), B2]), ComplexSpecies([A2, B2])])` Notice that ScaffoldSpecies can access their SubSpecies easily. The proposed indexing probably needs to be smarter for cases where there are multiple copies of the same SubSpecies in a ScaffoldSpecies - but that is doable, maybe by integer indexing.
`C4 = ComplexSpecies([A2, C2.species[B2])` ==> `Returns ScaffoldSpecies([ComplexSpecies([A1, B1]), B2]), ComplexSpecies([A2, B2])])` which is equal to `C3`.

Finally, if a subspecies is used in a reaction, it will be equivalent to using its containing ScaffoldSpecies.

For this to work, every ScaffoldSpecies will need a unique copy of each SubSpecies.

murrayrm:
Are the thoughts above intended to be part of this PR, or is that something that you are starting a discussion on (in which case you might want to start an issue?).

Note that there are other similar constructions that might occur in other places.  For example, if we create a protein fusion A-B then might have a. molecule that binds to just one of the proteins in the construct, to yield something like C:A-B or on the other protein to yield A-B:D.  Or a single protein might bind to both, so perhaps we write (C:A)-B, A-(B:D), and (A-B):E?

Similar things can happen with RNAs, genelets, kinases (eg, if there are multiple phosphorylation sites), methyltransferases, etc.

WilliamIX:
@murrayrm this pull request is a work in progress and active discussion. I think one advantage to having discussions around pull requests is that there is now commentary around the code changes in the request. With issues the code and discussion are frequently detached.

dr3y:
I don't really understand how ComplexSpecies would yield different things. I think that a function like make_complex() makes the most sense to me and overloading the ComplexSpecies constructor is too confusing.

I like the idea of passing a sub-member of a ScaffoldSpecies. Its basically another version of the binding site variable. The problem of duplicate SubSpecies is still present. The solution is the binding site variable.

The core problem of all of this is that if you have two ScaffoldSpecies that bind each other, we don't know how to handle that.

WilliamIX:
@dr3y responding to your comments below.

> I don't really understand how ComplexSpecies would yield different things. I think that a function like make_complex() makes the most sense to me and overloading the ComplexSpecies constructor is too confusing.

My hope here is that we can bury all the confusing aspects into the constructor and then use ScaffoldSpecies and SubSpecies the way we would use any other Species, but with the added bonus that they work differently when forming bound complexes and inside reactions. The issue I think people will be confused about in the future is when to use the binding site keyword, and when not to - the nice thing about the SubSpecies approach is that we basically make an object which takes care of that for us. I do agree however that writing this code will be complicated and overloading a constructor adds complexity to the code base, however I think it will make everything more consistent and object oriented.

> I like the idea of passing a sub-member of a ScaffoldSpecies. Its basically another version of the binding site variable. The problem of duplicate SubSpecies is still present. The solution is the binding site variable.

Instead of referencing a ScaffoldSpecies[SubSpecies] to get that SubSpecies, what if you could also reference ScaffoldSpecies[integer] to get that SubSpecies? This looks a lot like the binding site, variable in the end, but is just encapsulated into the species more directly.

> The core problem of all of this is that if you have two ScaffoldSpecies that bind each other, we don't know how to handle that.

This is a very good thought experiment I hadn't considered! We are definitely missing some pieces, but here are my first thoughts:

```
A1, A2, B1, B2= SubSpecies("A1"), SubSpecies("A2"), SubSpecies("B1"), SubSpecies("B2")

PA = ScaffoldSpecies ([A1, A2])
PB = ScaffoldSpecies ([B1, B2])
```

To make PA bind to PB with site A1 attaching to B1 we should be able to say very simply:

`C = ChemicalComplex([A1, B1])` 

However, I think @dr3y you correctly point out that it isn't obvious what this will return. Additionally, I worry that `C2 = ChemicalComplex([B1, A1])` will not end up being the same as `C`, which isn't the behavior we want. 

What this suggests to me is that we need a better internal Species representation than strings. I believe this problem has been dealt with by the DNA-strand-displacement using graph-isomorphisms. The increased level of complexity coming out of having binding sites like this also makes me even more convinced that we will need more than just a function create_complex in order to deal with the potential complexity arising when two things with diverse binding sites bind together. ScaffoldSpecies and SubSpecies might be a way to do this - although this example points out that a linear sequence representation is insufficient for two ScaffoldSpecies binding together. @zoltuz @ayush9pandey @murrayrm I would appreciate any feedback you have here - it seems like this issue is quickly becoming more complicated than I originally envisioned, but these are very plausible modelling scenarios biocrnpyler should be able to deal with.

dr3y:
> Instead of referencing a ScaffoldSpecies[SubSpecies] to get that SubSpecies, what if you could also reference ScaffoldSpecies[integer] to get that SubSpecies? This looks a lot like the binding site, variable in the end, but is just encapsulated into the species more directly.

So then each SubSpecies know what position it has in the ScaffoldSpecies? This is basically how DNA_construct and DNA_part works and that is why it allows having the same part in multiple locations.

WilliamIX:
> So then each SubSpecies know what position it has in the ScaffoldSpecies? This is basically how DNA_construct and DNA_part works and that is why it allows having the same part in multiple locations.

That sounds right - it makes sense for SubSpecies to functionally mimic DNA_part because a SubSpecies represents a DNA_part.

WilliamIX:
@dr3y some comments/notes on DNA_construct from the meeting:
1. DNAconstruct from a list of parts with no directions defaults to forward
2. reactions that understand OrderedPolymerSpecies & test cases
3. add get_part(part = part) which returns a part with the same name and type if one is inside. Please model this off of Mixture.get_component()
4. is it possible to save self.rnas, self.proteins so explore_txtl doesn't have to be called in update_reactions? These would have to be reset to None if anything is updated in the DNAconstruct. Same for out_components [this is especially important because this is an exponentially large operation, in general].
5.

biocrnpyler/components_basic.py
```diff
@@ -131,10 +131,11 @@ def __init__(
...
         for s in species:
             self.internal_species.append(self.set_species(s))
-        #bindee = self.internal_species[0]
-        #binder = self.internal_species[1:]
-        self.species = Complex(self.internal_species, name = name, material_type=material_type, attributes=attributes)
-
+        if(attributes is None):
+            attributes = []
+        self.species = Complex(species = self.internal_species, name = name, material_type=material_type, attributes=attributes)
+        print(self.species)
```
zoltuz, line 137: Can you remove this print?

biocrnpyler/dna_part_promoter.py
```diff
@@ -5,57 +5,101 @@
...
+class Promoter(DNA_part):
     """
     A basic Promoter class with no regulation. Needs to be included in a DNAassembly or DNAconstruct to function.
     """
     def __init__(self, name, assembly=None,
                  transcript=None, length=0,
-                 mechanisms=None, parameters=None, **keywords):
-        self.assembly = assembly
+                 mechanisms=None, parameters=None, protein=None,dna_to_bind = None, **keywords):
+        #self.assembly = assembly
```
zoltuz, line 20: please remove this line

PR #109, opened 2020-06-26
Minor SBML writing bug fix

ayush9pandey:
- added the required model ID when creating SBML model
- added name attribute as well to model and compartment when creating SBML model


zoltuz:
Screenshot 2020-06-26 at 11 07 57
There's a test which checks the content of the SBML file, which needs adjustment. (I guess the model ID is missing)

zoltuz:
btw, you do want this PR to be merged into master directly?

ayush9pandey:
Yeah,  I am looking into this, not sure how to fix this error.

ayush9pandey:
> 
> 
> btw, you do want this PR to be merged into master directly?

Yes, that would be helpful for the bootcamp (the overall changes are only in a couple lines). I didn't change the version number in order to avoid conflicts. 

Please excuse this negligence of the contribution rules.

ayush9pandey:
The source of the error is the following: I have added a random number to create a model ID. This random number changes every time and so test fails because it doesn't get the exact id in the SBML string. 

Suggestions on better ways to create a model ID that is not always the same for every model that is written out?

zoltuz:
> The source of the error is the following: I have added a random number to create a model ID. This random number changes every time and so test fails because it doesn't get the exact id in the SBML string.
> 
> Suggestions on better ways to create a model ID that is not always the same for every model that is written out?

How about generating that random number when you compile the CRN and saving it inside the CRN instance 
self._network_id =rand()

zoltuz:
> > btw, you do want this PR to be merged into master directly?
> 
> Yes, that would be helpful for the bootcamp (the overall changes are only in a couple lines). I didn't change the version number in order to avoid conflicts.
> 
> Please excuse this negligence of the contribution rules.

I see, no worries!

ayush9pandey:
> 
> 
> > The source of the error is the following: I have added a random number to create a model ID. This random number changes every time and so test fails because it doesn't get the exact id in the SBML string.
> > Suggestions on better ways to create a model ID that is not always the same for every model that is written out?
> 
> How about generating that random number when you compile the CRN and saving it inside the CRN instance
> self._network_id =rand()

I tried a different fix: Added an optional argument called `model_id` when creating SBML model so the model ID can be fixed to a desired name if needed. Changed the test code to have a fixed ID

ayush9pandey:
All green :) @zoltuz  can we merge now?

zoltuz:
> All green :) @zoltuz can we merge now?

Looks good! We can future proof it a bit better next time.

zoltuz:
Please merge these changes into the dev branch as well. Thanks!

ayush9pandey:
> Looks good! We can future proof it a bit better next time by extracting the model_id string as a constant.
Yeah, that would be better.

> Please merge these changes into the dev branch as well. Thanks!
Yes, will do. Should I create another PR for that?

PR #117, opened 2020-07-01
Bugfix mutable default arguments

zoltuz:
this PR fixes up the same bug ([mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)) scattered around the codebase.

Additionally, it changes the Travis config to execute a search for mutable default arguments in the code. If it finds one, fails the run. 

Please, **do not** squash the PR, it's been rebased already and I want to keep the bugfix and the Travis config change in separate commits. Thanks!


WilliamIX:
This PR also seems to run and makes sense to me. Will merge without squashing. I do have a question though: What does code like this do?

`species: List[Union[Species,str]]`

This kind of python is new to me.

zoltuz:
> This PR also seems to run and makes sense to me. Will merge without squashing. I do have a question though: What does code like this do?
> 
> `species: List[Union[Species,str]]`
These are [python type hints](https://realpython.com/lessons/type-hinting/). Right now, it doesn't do anything. (if you use VScode or Pycharm, then it show potential errors based on the type hints. Example:
Screenshot 2020-07-07 at 17 35 28

> 
> This kind of python is new to me. @zoltuz
I'm adding these in gradually, feel free to ignore them for now.

WilliamIX:
There were merge issues. To fix them, I this PR into my dev branch, got it working locally, then pushed dev.

Issue #115, opened 2020-06-29
True/False for leak isn't working


zjuradoq:
When working with example "4. Promoters, Transcriptional Regulation, and Gene Regulatory Networks.ipynb" I found that when I changed the leak to True or False the model was not effected. Image below shows result for two different values of leak when on or off.
![LeakCheck](https://user-images.githubusercontent.com/32440023/85962699-efd8ff80-b966-11ea-8c20-4b96cab344ac.JPG)

WilliamIX:
Thanks Zoila - I see the issue and it will be fixed in the next release. If this is a major issue for your work, I suggest you use RegulatedPromoter instead of Activatable/Repressible Promoter.

zoltuz:
Hi @zjuradoq, thanks for reporting this. This issue has been solved (thanks @WilliamIX! ) and the dev branch now contains the bugfix. 

If you want to give a try, you can install biocrnpyler using the code from the dev branch:
https://github.com/BuildACell/BioCRNPyler/wiki/How-to-install-BioCRNPyler#installing-the-dev-branch

Issue #92, opened 2020-06-11
Material type inheritance for complexes


mkratz:
When applying a filter for dilution and degredation, it seems that ComplexSpecies() objects will inherit the material types of both their parent species. 

For instance, if we create an object, ComplexSpecies([RBZ,mRNA])  and set dilution to True for RBZ, mRNA and set mRNA degredation to True, the resulting Complex will be both degraded and diluted. If we look at the source code (global_mechanism.py) we see why, for applying dilution and degredation for ComplexSpecies objects, apply_filter() will look at the species that make up the object.  Further, this results in conflicting degredation conditions results in an error e.g. Complex to False and RBZ or mRNA to True. 

Ideally, this would be a decision the user could make on their own i.e. should dilution be applied based on ONLY complex material_type or based of all constituent material types through some kwarg. 

I've attached a parameter txt and a code txt (just rename file extension to ipynb) which demonstrates this issue. 
[GHI_Demo.txt](https://github.com/BuildACell/BioCRNPyler/files/4762085/GHI_Demo.txt)
[params.txt](https://github.com/BuildACell/BioCRNPyler/files/4762086/params.txt)

zoltuz:
I could reproduce the error based on the example file. Thanks for sending it @mkratz

WilliamIX:
This will be fixed in my newest PR (which initially will go to the dev branch) and then be fixed in the next merging of dev into master at the end of the month.

zoltuz:
I've checked the example attached to this issue with the code in #119, it no longer gives an error.

zoltuz:
Hi @mkratz,  thanks for reporting this. This issue has been solved (thanks @WilliamIX! ) and the dev branch now contains the bugfix.

If you want to give a try, you can install biocrnpyler using the code from the dev branch:
https://github.com/BuildACell/BioCRNPyler/wiki/How-to-install-BioCRNPyler#installing-the-dev-branch

PR #103, opened 2020-06-20
new object based Propensities

zoltuz:
Checklist

- [X] rename k_forward to k in Jupyter notebooks and in other places  
- [X] add backward compatibility to Reaction: ```Reaction(inputs=[sp1],output=[sp2],k=0.1)``` and ```Reaction(inputs=[sp1],outputs=[sp2],propensity_type=HillNegative(...))``` should work
- [X] add unit tests for non-massaction kinetic types
- [X] add reverse reaction setup in sbmlutil
- [X] propagate propensity_annotation (rename it! see: #127) to kinetic_law from the user interface

The idea of this feature is to make the Propensity types object-based, which supports better type checking and more maintainable code. 
Please have look and let me know what you think!

usage:

### massaction kinetic
```python
mak = Massaction(k_forward=0.15)
s_a = Species('a')
s_b = Species('b')
rxn = Reaction(inputs= [s_a], outputs=[s_b], propensity_type=mak) 
```
it also has a convenient way to initialize a Reaction object
```python 
rxn = Reaction.from_massaction(inputs=[s_a], outputs=[s_b], k_forward=0.15)
```
### Hill posititve
```python
s1 = Species('s1')
hp = HillPositive(k=0.2, K=50, s1 = s1, n=3)
s_a = Species('a')
s_b = Species('b')
rxn = Reaction(inputs= [s_a], outputs=[s_b], propensity_type=hp) 
```
It also has type checking under the hood:
```python
s1 = Species('s1')
s_a = Species('a')
s_b = Species('b')
rxn = Reaction(inputs= [s_a], outputs=[s_b], propensity_type=s1) 
-> raise Type error unknow propensity
```

## Internal changes
A new data structure was created: 
```python
@dataclass(frozen=True)
class WeightedSpecies:
    species: Species
    stoichiometry: int = 1
```
why?
```python
rxn.input_coeff.pop() # this just kills the whole CRN
```
This changes how reactions containing species with higher than one stoichiometry is created:
```python
sp1 = Species(name='s1')
sp2 = Species(name='s2')  
cp_s = ComplexSpecies([sp1,sp2])
inputs = 3*sp1+5*sp2
output = 2*cp_s
rxn = Reaction.from_mass_action(inputs=inputs, outputs=output,k_forward=0.2)
```
everything with ```stoichiometry=1``` is converted under the hood:
```python
sp1 = Species(name='s1')
sp2 = Species(name='s2')  
[...]

PR #119, opened 2020-07-09
Minor Bug Fixes

WilliamIX:
This pull request fixes a some issues on the issues page including:

https://github.com/BuildACell/BioCRNPyler/issues/115 (fixing leak in Repressible and Activatable Promoter classes)

https://github.com/BuildACell/BioCRNPyler/issues/92 (fixing filter_dict issues in GlobalMechanisms). New tests and documentation were added here as well.

And making sure build_crns_directly.xml is ignored.


biocrnpyler/chemical_reaction_network.py
```diff
@@ -232,6 +236,18 @@ def replace_species(self, species: Species, new_species: Species):
         
         return ComplexSpecies(species = new_species_list, name = new_name, material_type = self.material_type, attributes = self.attributes)
 
+    #Returns all species in the ComplexSpecies. If recursive = True, returns species inside internal ComplexSpecies recursively as well.
+    def get_species(self, recursive = False):
```
zoltuz, line 240: Is there a strong case for recursive=True? I plan to change this to a more pythonic code in #103, namely:
```python
class ComplexSpecies:
     @property 
     def species(self):
         return self.species
```    
Then, you can use it like:
```python
cp = ComplexSpecies([sp1,sp2])
print(cp.species)
```
You don't need to change the code, that's not my point, just I want to know why recursive=True is needed right now.

biocrnpyler/dna_assembly_promoter.py
```diff
@@ -108,11 +108,13 @@ def update_reactions(self):
 
 #A class for a promoter which can be activated by a single species, modelled as a positive hill function
 class ActivatablePromotor(Promoter):
-    def __init__(self, name, activator, transcript = None, **keywords):
+    def __init__(self, name, activator, leak = False, transcript = None, **keywords):
```
zoltuz, line 111: This is fine for now, but this change breaks all the code which use the old format with positional arguments:
```python
ap = ActivatablePromotor(name,activator,transcript)
```
I guess this class is not used extensively right now, so it's fine for now

biocrnpyler/dna_assembly_promoter.py
```diff
@@ -125,25 +127,31 @@ def update_species(self, **keywords):
...
+
         reactions += mech_tx.update_reactions(dna = self.assembly.dna, transcript = self.transcript, regulator = self.activator, 
-                                             component = self, part_id = self.name+"_"+self.activator.name, **keywords)
+                                             component = self, part_id = self.name+"_"+self.activator.name, leak = self.leak, **keywords)
+
+        
         return reactions
 
 #A class for a promoter which can be repressed by a single species, modelled as a negative hill function
 class RepressablePromotor(Promoter):
```
zoltuz, line 147: RepressiblePromoter

biocrnpyler/dna_assembly_promoter.py
```diff
@@ -108,11 +108,13 @@ def update_reactions(self):
 
 #A class for a promoter which can be activated by a single species, modelled as a positive hill function
 class ActivatablePromotor(Promoter):
```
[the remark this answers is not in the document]
WilliamIX: whoops that would be me. Will change.

biocrnpyler/chemical_reaction_network.py
WilliamIX: So I think this is a little different than just accessing ComplexSpecies.species.

In the non-recrusive case, only the ComplexSpecies is returned. In the recursive case, all the Species in the ComplexSpecies are returned and furthermore, if the Species inside the ComplexSpecies is also a ComplexSpecies, the species inside that ComplexSpecies are returned and so on recursively.

I think this actually makes 3 cases, which maybe we can have different functions for (or the same function with different keywords):

1. Get the list of species inside a ComplexSpecies, with no kind of recursion.
2. Get the list of species inside a ComplexSpecies with recursion (ie flatten all nested ComplexSpecies).
3. A base-case for 2 so when a Species is hit the recursive function stops.

In terms of this code, it is important to have both recursive and non-recursive calls in order to filter ComplexSpecies correctly in different cases when using GlobalMechanisms. I can also imagine other cases when we might want to flatten all nested ComplexSpecies.

biocrnpyler/dna_assembly_promoter.py
WilliamIX: I will fix this too - good catch thanks!

biocrnpyler/dna_assembly_promoter.py
```diff
@@ -125,25 +127,31 @@ def update_species(self, **keywords):
...
-                                             component = self, part_id = self.name+"_"+self.activator.name, **keywords)
+                                             component = self, part_id = self.name+"_"+self.activator.name, leak = self.leak, **keywords)
+
+        
         return reactions
 
 #A class for a promoter which can be repressed by a single species, modelled as a negative hill function
-class RepressablePromotor(Promoter):
-    def __init__(self, name, repressor, transcript = None, **keywords):
+class RepressablePromoter(Promoter):
```
zoltuz, line 147: Sorry for being a grammar nazi, but I think it is called a RepressiblePromoter
e.g. https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3599224/

WilliamIX: Sorry for being a bad speller. Typo-corrected.

zoltuz: no worries, just this PR it the perfect place to weed out this typo.  Thanks for taking care of it!

PR #121, opened 2020-07-10
changing pytest config

zoltuz:
Just a tiny PR, changing the pytest config a bit

adding verbose test output (-vv), colors (--color=yes) and adding some ignores for some warnings (filterwarnings)


WilliamIX:
This all works on my computer. Squashing and Merging!

PR #124, opened 2020-07-14
New parameter classes

WilliamIX:
This PR adds new classes to deal with parameters.

## Main new classes are:

- ParameterDatabase: Stores ParameterEntries and looks them up based on keywords. Contains useful functionality like "in", indexing, iteration, etc. Currently ParameterDatabase enforces that all parameter keys be  3-tuples (mechanism, part_id, parameter_name) [None is allowed] or just a string parameter name. Also parses and loads parameters from files, dictionaries and other ParameterDatabases.

- Parameter: High level class to store a value and a name

- ParameterEntry(Parameter): Stores a value, name, arbitrary keys such as "mechanism":"mechanism name" in case we want more complex parameter keys in the future. Also stores a generic info dictionary which takes other columns from a parameter file.

- ModelParameter(ParameterEntry): a ParameterEntry with additional information about the key used to search for the ModelParameter and the key the ModelParameter was found under after defaulting. This data will be used to display where parameters come from and name them in SBML.

## Changes to core code:
There are numerous changes to the core code and examples in order to integrate these new classes. A summary is:

- Mixture.get_parameter(...) and component.get_parameter(...) were rewritten to use the new classes. In particular, Components now store the Mixture they are in and call functions from Mixture when looking for parameters.

- Mixture.set_initial_condition(...) and component.get_initial_condition(...) were rewritten to use the new classes.

- Unit tests for Parameter classes and changes to minor changes to previous unit tests and ipynb's as required to make everything work.

- Combinatorial testing of Components and Mixtures. Very helpful in finding bugs and increasing coverage! Many bugs were fixed in this process. 

- There were also some seemingly unrelated changes to a variety of TxTl mechanisms - in particular for all transcription mechanisms the protein keyword is always passed in. Importantly, if protein = Species.. and transcript = None, all transcription mechanisms now produce the protein Species directly. This allows for ExpressionExtracts to work without any Species replacement many more Transcription Mechanisms. This change was necessary with this PR because of changes to how compile_crn worked which were incompatable with Species replacement (which, was kind of hacky anyways to be honest). Also added unit tests of Species, Reaction and ChemicalReactionNetwork .replace species which were no longer covered by testing after these changes.

## Future considerations and things this PR doesn't do:

- This PR does not integrate new Parameters with reactions. I was waiting till the new reaction classes are out to do this. Major change will be returning the ModelParameter from Component.get_parameter(...) instead of the ModelParameter.value (which is what is currently happening). Once integrated with Reactions, we should be able to easily print the Parameters used in Reactions and also save parameter names as global SBML parameters based upon their keyword tuples.

- This PR should provide better ways to print parameters in reactions - but this is closely linked to the reaction.pretty_print and should be included after the new Reactions are out.

- parameter_warning keyword has been turned off. I believe with the new parameter system, where it is will be easy to print a CRN and get parameter information, these excessive warning messages are no longer useful. However, keyword is still in place (but not functional currently) until we officially decide to remove it.

- I believe the way parameters work is much more elegant than the previous dictionaries and we should consider using a similar system for Mechanisms. This will be much more extendable and may solve some of the issues @zoltuz has previously raised with how Mechanisms work.


zoltuz:
Some comments:

### using namedtuple instead of a tuple
There're two built-in datatypes that you can use instead of the tuples you have in the code:
https://pymotw.com/2/collections/namedtuple.html
https://realpython.com/python-data-classes/ 

### replacing comments with docstrings:
You've added some comments here and there in the code, you could convert some of them to docstrings with minimal effort.

WilliamIX:
> ### using namedtuple instead of a tuple
> ### replacing comments with docstrings:

Both of these have been done in my most recent commits.

zoltuz:
Hi @WilliamIX 
There were two inactive tests in test_parameter.py (test_ prefix was missing). Now those two tests are also executed, but they are failing, please have a look. 

I commented out some values for now:
Screenshot 2020-07-18 at 20 15 57
Please do NOT reformat the example_csv variable on line 115, otherwise, the test won't work. No idea why feel free to explore it, but I remember I spent half-day playing with it last year until I got this one right.

Tests/test_parameter.py
```diff
@@ -3,93 +3,276 @@
...
+
+    def test_get_field_names(self):
+        PD = ParameterDatabase
         test_accepted_field_names = {'mechanism': ['mechanism', 'mechanism_id']}
         valid_field_names = ['part_id']
 
         # test None as field_names
         with self.assertRaisesRegexp(ValueError, 'field_names must be a list of strings'):
-            Parameter._get_field_names(field_names=None, accepted_field_names=test_accepted_field_names)
+            PD._get_field_names(field_names=None, accepted_field_names=test_accepted_field_names)
```
zoltuz, line 70: I'm not sure why you call a static method through the object (PD) and not through the class (Parameter) as before. Is there a reason for that?

biocrnpyler/mixture.py
```diff
@@ -134,75 +130,93 @@ def add_components(self, components: Union[List[Component],Component]):
             assert isinstance(component, Component), \
                 "the object: %s passed into mixture as component must be of the class Component" % str(component)
             self.components.append(component)
+
+            #Reset components Mixtures
+            component.set_mixture(self)
+
             component.update_mechanisms(mixture_mechanisms=self.mechanisms, overwrite_custom_mechanisms = False)
-            component.update_parameters(mixture_parameters=self.parameters)
+            #component.update_parameters(mixture_parameters=self.parameters)
```
zoltuz, line 138: do we need this line anymore?

biocrnpyler/parameter.py
```diff
@@ -64,103 +45,319 @@
...
+    """
+    def __init__(self, parameter_name: str, parameter_value: Union[str,numbers.Real], parameter_key=None, parameter_info=None):
+        Parameter.__init__(self, parameter_name, parameter_value)
+
+        self.parameter_key = parameter_key
+        self.parameter_info = parameter_info
+
+    #Helper function to create ParameterKeys
+    @staticmethod
+    def create_parameter_key(new_key: Union[Dict, ParameterKey, str], parameter_name=None) -> ParameterKey:
```
zoltuz, line 125: should be this method private? i.e adding an underscore at the beginning of the method?
```python
def _create_parameter_key(new_key: Union[Dict, ParameterKey, str], parameter_name=None) -> ParameterKey:
```

biocrnpyler/parameter.py
```diff
@@ -64,103 +45,319 @@
...
+
+    @search_key.setter
+    def search_key(self, search_key):
+        self._search_key = self.create_parameter_key(search_key, self.parameter_name)
+
+    @property
+    def found_key(self):
+        return self._found_key
+
+    @search_key.setter
```
zoltuz, line 194: this is a typo!

biocrnpyler/parameter.py
```diff
@@ -64,103 +45,319 @@
...
+            #Used for flexible column headings
+            accepted_field_names = {
+                'mechanism': ['mechanism', 'mechanism_id'],
+                'param_name': ["parameter_name", "parameter", "param", "param_name"],
+                'part_id': ['part_id', 'part'],
+                'param_val': ["val", "value", "param_val", "parameter_value"]
+            }
 
-            field_names = Parameter._get_field_names(csvreader.fieldnames, accepted_field_names)
+            field_names = self._get_field_names(csvreader.fieldnames, accepted_field_names)
```
zoltuz, line 356: again, is there a reason why this method is called through the object rather than the class?

zoltuz:
> > ### using namedtuple instead of a tuple
> > ### replacing comments with docstrings:
> 
> Both of these have been done in my most recent commits.

Thanks! I was referring to the comments in the new files related to the Parameters not all the files in the PR. Anyhow thanks for the effort!

Tests/test_parameter.py
WilliamIX: In my mind, ParameterDatabase loads files, Parameter just stores the data.

biocrnpyler/mixture.py
WilliamIX: I don't think so - good catch.

biocrnpyler/parameter.py
WilliamIX: I could imagine someone wanting to use the method - for example to convert parameter values in a dictionary or another program to biocrnpyler parameters.

biocrnpyler/parameter.py
WilliamIX: get_fields_names had been a class method before and I kept it that way - in some sense, it isn't linked to the class in any way - it doesn't modify any class variables - so it works to use it as a class method. If you think it should be an instance method though, I have no strong opinions either way.

biocrnpyler/parameter.py
WilliamIX: fixed.

WilliamIX:
@zoltuz I think I fixed all the problems you mentioned. Removed one test which was failing (couldn't get it to consistently load relative file paths). It was redundant with your "pretend to make a file via code" solution so I think that is okay.

PR #126, opened 2020-07-16
Ordered polymer species

dr3y:
Updates to Species logic which allow you to have:

1) OrderedPolymers which bind to things through one of their monomers, and have positions and directions

2) Species that have a position and direction (OrderedMonomer)

3) Complex() function which automatically decides whether to make a normal ComplexSpecies or a OrderedPolymerSpecies

WARNING: This is a draft which does not include tests or good examples yet. Code was ripped from DNA_construct where many of these OrderedPolymer things are used.


dr3y:
About `OrderedPolymerSpecies`:

`species_set` is defined for `ComplexSpecies` so it should also work here

`self._polymer` is made by `OrderedPolymer` and is the same as `species` from `OrderedComplexSpecies`

`BaseSpecies` is the unbound version of the polymer

WilliamIX:
Generally speaking, I think this looks good. I'm playing around with these Species in a jupyter notebook and here are some suggestions:
1. OrderedMonomer and OrderedPolymer aren't really part of CRN so they should be in their own file, maybe polymer.py.
2. When instantiating OrdereredPolymer with something that isn't a list, an error should be thrown.
3. I would make OrderedMonomer much more minimal: get rid of repr, hash .data, pretty_print, show_bindloc, contains. Let Subclasses deal with that and make OrderedMonomer do the bare minimum.
4. Similarly for OrderedPolymer - get rid of everything you don't need noting that the Species Subclasses and DNASubclasses will have to overwrite many of these methods anyways. I think conceptually, these two classes are made to be Subclassed.
5. In OrderedPolymerSpecies, a lot of the constructor is just duplicated from OrderedComplexSpecies (like self.initial_conc etc) maybe just reuse that?
6. Will OrderedPolymerSpecies._polymer always return the same thing as OrderedPolymerSpecies.species?
7. when passing in a non-species to OrderedPolymerSpecies constructor, the assertion error that comes out provides no error message.
8. Where are you using the make_species helper function?
9. When do you use OrderedPolymerSpecies.base_species? 

Thoughts on the class Complex and creating ComplexSpecies more intelligently:
1. Make Species, ComplexSpecies, and OrderedComplexSpecies and OrderedPolymerSpecies constructors all be the same - namely what Species currently has. This will in effect create "Empty" species lists but symbolically, these classes are still formal CRN Species.
2. Have class_methods for ComplexSpecies, OrderedComplexSpecies, and OrderedPolymerSpecies .from_list(list_of_species) which automatically create these classes from lists of Species. Importantly, if there is an Species with a Parent (eg a Species that is part of an OrderedPolymerSpecies) you can do any logic needed here.
3. If a species_list is passed into a ComplexSpecies constructor, raise an error that suggests creating the species using the .from_list class method.

dr3y:
1. they are the most essential part of the CRN?
2. fixed
3. fixed
4. i think it is minimial? A lot of methods are required to get list functionality
5. OrderedPolymerSpecies generates its own name, and sets the base_species. Also, ComplexSpecies allows strings to be put in, which OrderedPolymerSpecies doesn't. I think they are different enough.
6. yes. But before you subclass OrderedPolymer, the contents arent necessarily species? Thus, the conversion via the property.
7. fixed
8. I wanted to use it whenever you use set_species. However, we also discussed that anything besides the Species constructor should take Species, so I guess it isn't used? Maybe it's not useful
9. This is used in DNA_construct in order to find DNAs which have other things bound to them
10. yes this is because S1 is a persistent reference to that species. If the species is now part of a different polymer then it knows it is part of a different polymer. I think what you are saying would be compatible and would work. I don't have a strong opinion i guess except for the fact that I don't want to change what works.
11. so it works now, but its still kinda weird I think.

1.-3. My personal opinion here is that doing what you say will be a more complicated way of just using "Complex()" and we will have to duplicate code in all these class methods which will be equivalent to just using "Complex()" anyway.

WilliamIX:
Thanks Andrey! Just replying to your comments below. I'm going to make some of these changes myself but thought I'd document my thinking.

> 1. they are the most essential part of the CRN?

You also use them for DNA_parts, so know, they are a level of abstract parallel to and under ChemicalReactionNetwork. I made this change on one of my branches and it didn't break anything - and so long as Components can also be OrderedMonomers and OrderedPolymers I think we should make it a separate file. I could also imagine a more abstract version which represents a graph that could be used as either a Species or Component subclass in the future.

> 6. yes. But before you subclass OrderedPolymer, the contents arent necessarily species? Thus, the conversion via the property.

Oh I see - that is a great solution!

> 8. I wanted to use it whenever you use set_species. However, we also discussed that anything besides the Species constructor should take Species, so I guess it isn't used? Maybe it's not useful

I will go ahead and delete this in my branch then unless you object.

> 10. yes this is because S1 is a persistent reference to that species. If the species is now part of a different polymer then it knows it is part of a different polymer. I think what you are saying would be compatible and would work. I don't have a strong opinion i guess except for the fact that I don't want to change what works.

It needs to be possible to add Species to multiple Polymers - I think the best implementation is via copying. I will take a stab at this myself as the issue is intricately related to synchronizing Species at Compile time which I am planning on coding up as well.

> 11. so it works now, but its still kinda weird I think.

I agree, for now it is weird and we should throw an Error if this happens. Will take add myself.

> 1.-3. My personal opinion here is that doing what you say will be a more complicated way of just using "Complex()" and we will have to duplicate code in all these class methods which will be equivalent to just using "Complex()" anyway.

I think we should continue this discussion tomorrow and come to a firm conclusion. I worry about using a Complex() function for two reasons: 1) how do we differentiate between OrderdComplexSpecies and ComplexSpecies and what if more classes are added in the future? and 2) how do we enforce that people use this function so code isn't accidentally not compatible? I was thinking that using a class function .from_list would solve this problem (and using inheritance we can actually reuse the code I believe). However I am likely missing some key points because I haven't worked as closely on this as you have.

WilliamIX:
I think we are very very very close. I have a couple of edge case worries about the Complex() function I am documenting here. I will take a stab at them myself.

1. Complex() only works for ComplexSpecies, not OrderedComplexSpecies. I believe both are needed.
2. Complex() is hard to follow, but very important to the functioning of the code - a doc string and better comments would be helpful.
3. I think we should have a keyword so ComplexSpecies and OrderedComplexSpecies throw a warning if not called by the Complex() function. Initially, we can even make this an error to make sure we make all the relevant changes in the code so everything becomes compatible.
4. I am unclear on why valent_complex.material_type is reset to "ordered_polymer".

Issue #129, opened 2020-07-25
Add SBML validation


ayush9pandey:
add sbml validator that automates http://sbml.org/Facilities/Validator/ for BioCRNpyler generated models so that we can be sure that the SBML output from biocrnpyler is globally readable by any SBML reading tool.

zoltuz:
PR #131 solved this

Issue #128, opened 2020-07-24
Things to do before we merge the current dev into master


zoltuz:
This is a placeholder issue to collect all that needs to be done before we merge dev into master.
Feel free to add items to the list!

- [X] DNA Construct PR (Andrey #102 )
- [X] DNA construct numpy/matplotlib bug #146 
- [x] refactor the SBML interface, see #122 (Ayush PR)
- [x] test MAK and other non-massaction kinetic types via bioscrape (Ayush PR)
- [x] test MAK and other non-massaction kinetic types via libroadrunner (Ayush PR)
- [x] test MAK and other non-massaction kinetic types via Copasi [@zoltuz] (Ayush PR)
- [x] test stochastic MAK simulation via bioscrape  (Ayush PR)
- [X] update version information for biocrnpyler (Version 1.0.0!) #140 
- [x] Add copying for Components placed into Mixtures and appropriate getters to retrieve internal Components. see #123 (William PR)
- [X] Add copying for Species & Reactions placed into CRNs and appropriate getters to retrieve internal Species. This step should simultaneously ensure that duplicate species are merged. see #120 (William CRN PR TODO)
- [X] Add jupyter documentation to Scope (above two changes). (William CRN PR TODO)
- [X] Add jupyter documentation and more unit tests to OrderedPolymerSpecies class. In particular documentation & tests of how this class interacts with ComplexSpecies vs Reactions is crucial (William CRN PR TODO)
- [x] Add jupyter documentation to viewing found / search parameters used in CRN compilation (William PR)
- [X] create Contributing.md see #107 (Zoltan)
- [X] edit README.md, welcome message, how to install, where to start (William & Zoltan)
- [X] register the package at PyPi and upload it (create a git tag and package with matching version number) (Ayush + Zoltan)
- [x] Create SBtools Developer Mailing List for BioCRNpyler (Ayush)
- [X] Add more backwards compatibility for old propensity interface. Currently, only Massaction is supported. (Zoltan in #141 )
- [x] Hoverboxes over reaction lines in plotting are no longer displaying anything. (Andrey)
- [X] get rid of all parameter warnings. These are no longer necessary now that parameters can be displayed with pretty_print. (William PR #134 )
- [X] Update Wiki /README (Zoltan)
- [X] Read the Docs! (Ayush) ZAT: Richard set up the webhook for us. 
[...]

Issue #127, opened 2020-07-20
Make SBML writing annotations for bioscrape toggle-able


ayush9pandey:
Add 'for_bioscrape' optional keyword argument to `write_sbml_file` function with default False to not write the propensity annotations always.

murrayrm:
I was going to mention that `for_bioscrape` seemed a bit clunky as a keyword.  Given that we may eventually have multiple package-specific annotations for BioCRNpyler functions, perhaps something more general (including even a list) would make sense?  Can be more general than just for SBML output.

ayush9pandey:
Yes, that's a helpful suggestion, will take that into account.

zoltuz:
this feature was added in #103, thanks for suggesting this feature!

ayush9pandey:
Reopening to slightly refactor how it is set up right now after #103. I am changing the behavior a bit so that the higher level functions only see a **kwargs and the `create_kinetic_law` function also remains general in the sense that it doesn't have bioscrape specific things in it. It will call a `create_annotations` function with the **kwargs passed in so that this functionality can be extended in the future. Let me know if you have any thoughts...I will try to make it work this way.

zoltuz:
I agree that's the best way to handle it. In the many merges recently, somehow,  for_bioscrape argument sneaked out of kwargs, it's good to put it back.

ayush9pandey:
Yes, I think it's not out in the open anymore - always carried through kwargs (except in the some of the test code). But due to this move I think something very minor is breaking in #131 ..

zoltuz:
I think PR #131 now solved this

Issue #122, opened 2020-07-11
Add species modifier reference to species used in reaction rate laws that are neither reactants or products


ayush9pandey:
SBML needs all species referred in a reaction or a kinetic law to have an associated species reference object. I had fixed this issue in bioscrape/sbmlutil.py, so would just need to port those changes over to biocrnpyler so that valid SBML models are created. Currently, the SBML models are not valid whenever hill propensities are used as shown here: 
![image](https://user-images.githubusercontent.com/8139195/87211914-c312d980-c2d0-11ea-97e2-96e8c580a86f.png)

COPASI could not import biocrnpyler generated model due to this error. I will fix this  but creating this issue to track it.

zoltuz:
PR #131 solved this

PR #131, opened 2020-07-25
SBML refactoring and bug fixes

ayush9pandey:
To address issues in:-
* #129 
* #122 
* #128 
* #127


zoltuz:
There's a bug, which came from the previous PR, sorry about that! 
Screenshot 2020-07-26 at 12 44 42
There are two options to solve this:
1, use defaultdict, which creates the missing keys in the dictionary on the fly
2, make a copy of ```self.propensity_dict```, just like here:
https://github.com/BuildACell/BioCRNPyler/blob/dev/biocrnpyler/propensities.py#L401

ayush9pandey:
Thanks, Zoltan. i used the proposed solution # 2 and it did fix some of the errors but I am still seeing the key error 'k' at a few places. I will look at how to fix that. 

#122 should be fixed now, also #127 (once the above error is fixed). 

Next commit will do #129.

ayush9pandey:
Opening it for review. The code changes required are close to done..only need to finish up with bug fixes so that all the tests run.

zoltuz:
Hurray! All test passed. You want to add more for the PR @ayush9pandey  or it is ready for review?

ayush9pandey:
The coverage went down I guess because I added the modifier reference code and that is not being tested. Also, the validate SBML functions are being used for testing but are not tested themselves so probably causing the lowered coverage percentage. I won't bother with this for now but we might wanna add simple tests for these things or move them to more appropriate places in the code. 

I will see if I can fix #135 as well with this PR @zoltuz  before requesting review.

ayush9pandey:
General propensity code was also not being tested earlier, so with this next commit I have added some refactored testing code that does that. The bioscrape annotation testing code is now in a new function which makes more sense IMO. 

This commit will probably break things again but should be easy to fix those.

zoltuz:
Hi @ayush9pandey 
Please a write_to_sbml and validate_sbml to https://github.com/BuildACell/BioCRNPyler/blob/dev/Tests/test_mixtures_and_components_combinatorially.py

biocrnpyler/propensities.py
```diff
@@ -167,47 +167,122 @@ def from_dict(cls, propensity_dict):
         merged.update(propensity_dict['species'])
         return cls(**merged)
 
-    def _create_bioscrape_annotation(self, model, propensity_dict_in_sbml):
+    def _create_annotation(self, model, propensity_dict_in_sbml, **kwargs):
+        '''
```
zoltuz, line 171: Usually, triple double-quote is used for docstring: """, but you can leave it like this if you want.

biocrnpyler/sbmlutil.py
```diff
@@ -225,18 +223,28 @@ def _create_reactants(reactant_list, sbml_reaction, model):
         # What to do when there are multiple species with same name?
```
zoltuz, line 223: Do we want to remove this comment?

biocrnpyler/sbmlutil.py
```diff
@@ -419,3 +426,94 @@ def getSpeciesByName(model, name, compartment=''):
...
+
+def validate_sbml(sbml_document, enable_unit_check = False, print_results = True):
+    """
+    Validates the generated SBML model by using libSBML SBML validation code
+    """
+    validator = validateSBML(enable_unit_check)
+    validation_result = validator.validate(sbml_document, print_results = print_results)
+    if validation_result > 0:
+        warn('SBML model invalid. Run with print_results = False to hide print statements')
+    return validation_result
```
zoltuz, line 519: add a new line at the end. pls

biocrnpyler/sbmlutil.py
ayush9pandey: Did we do something about this? Is it printed out as an error/warning some place?

biocrnpyler/propensities.py
ayush9pandey: Fixed.

biocrnpyler/sbmlutil.py
ayush9pandey: Done.

ayush9pandey:
I am working on updating the notebooks so that they all work with the current changes in dev...these are minor things that are causing errors..we can merge the PR once I have checked all the notebooks are working, generating simulation plots, and the SBML files are valid + import-able by other tools (COPASI).

ayush9pandey:
I think this PR is ready to be merged from my end. @WilliamIX  and @zoltuz  please feel free to suggest changes (particularly look at the couple lines here : https://github.com/ayush9pandey/BioCRNPyler/blob/f178458e8c6820e42fd1c8ff709888530c1d2aeb/biocrnpyler/propensities.py#L211 ). With this change - replacing the first appearance of "k_forward" by "k" fixes the bioscrape annotation problem that we discussed today. It is ofcourse hacky, but I think a lot of things with bioscrape annotations are that way. In case you have a better way to solve this or if you think this will break some use case let me know.

ayush9pandey:
Also, @zoltuz  any idea why the codecov/project check failed? That one always seems to fail for me :(

WilliamIX:
> Also, @zoltuz any idea why the codecov/project check failed? That one always seems to fail for me :(

The issue is that you are losing .35% of testing coverage overall. It looks like this is mostly happening in propensities.py and sbmlutil.py - largely this is due to the new validate_sbml function, which is used for testing, so I think this is ok. If you want to pass though, add some more unit tests to cover other parts of the code.

ayush9pandey:
All green! New tests added for SBML validation and libsbml validation.

biocrnpyler/sbmlutil.py
WilliamIX: I think we should definitely have a warning

PR #141, opened 2020-07-30
adding full compatibility with the old Reaction class interface

zoltuz:
Reaction now workings with Propensity types, but for compatibility reasons, the old (string-based) is supported for now.

- [X] support for all non-massaction types
- [X] unit tests with the old interface format


WilliamIX:
Code and tests look good and it all works on my computer as well. Will merge

Issue #146, opened 2020-08-01
DNA construct has introduced a dependancy on matplotlib


zoltuz:
Hi @dr3y ,

Somehow, I missed that during the review but dna_construct explicitly requires matplotlib (and numpy as well) inside the core code. I have removed numpy already in my version to make the core code numpy (and matplotlib) free, now all the dna_constrct test are broken. 

Can you remove matplotlib or put inside a plotting related function asap? Many thanks!

dr3y:
hi i just submitted a pr #149 to addres this

zoltuz:
resolved in PR #149

PR #147, opened 2020-08-01
factor out numpy dependency

zoltuz:
Let's wait until #146 gets resolved then we can merge this


ayush9pandey:
Merging this PR now. The coverage went down quite a bit because of all the plotting testing that is not going through due to optional dependency refactor changes.

PR #152, opened 2020-08-01
add docstrings and some PEP8 formatting

zoltuz:
it also bugfix some typos and the length setter in component.py


zoltuz:
> Didn't read all the docstrings but seems like this is ok to be merged!

I know, lots of lines changes, but it should be fine.

PR #159, opened 2020-08-02
Dna construct examples fix

dr3y:
I've fixed a few typos in dna construct examples which were causing errors. For instance we had `import matplotlib as plt` which should be `matplotlib.pyplot as plt`, and some plotting commands were still commented out

I did this by merging master with my master and then creating a branch so im not sure if this is the best way to merge these changes, but i'm going to try


ayush9pandey:
I am not sure why there's this long list of commits when the real commit change is only the last one.

Let's keep our workflow intact and have this PR to the dev branch? Because I also have some minor things that I need to merge so I will also do that on dev first then we can merge to master later? Who knows maybe that will also fix the problem with all these commits appearing here.

zoltuz:
Someone our individual forks still remembers all the commits we have squashed. When we open the pull requests, all of these already squash commits show up, but in the Files change tab the correct difference is shown.

dr3y:
i am testing this PR in my computer because travis does not check plotting based examples any more

dr3y:
well it seems to pass fine for me, 81% coverage is what my computer says

zoltuz:
Cool! I suggest merging this to dev first, I'm sure other issues will come up to today. We can bundle all of them together and push to master later on.

zoltuz:
Looks good to me.

PR #164, opened 2020-08-03
Actually support old dnaplotlib

dr3y:
added try except to renderDNA call


dr3y:
dont mind the commit number, look at the changed files

PR #166, opened 2020-08-03
Binder runnable

dr3y:
i added requirements.txt and i put in the try except so that the default version of dnaplotlib can be used


requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+fa2
+lxml
+beautifulsoup4
+cython
+sympy
+pandas
+astropy
+emcee
+python-libsbml
+#--editable git+git://github.com/biocircuits/bioscrape.git@master#egg=bioscrape
```
zoltuz, line 14: do you want to remove this for now?

dr3y: ok let me remove it and make sure the binder still works. it should though

dr3y: i tried to make it load bioscrape obviously that didnt work

zoltuz: I figured, but let's not accumulate dead code.

requirements.txt
```diff
@@ -0,0 +1,10 @@
+--editable git+git://github.com/dr3y/dnaplotlib.git@master#egg=dnaplotlib
+bokeh
+matplotlib
+networkx
+fa2
+lxml
+beautifulsoup4
```
zoltuz: do we need this?

requirements.txt
```diff
@@ -0,0 +1,10 @@
+--editable git+git://github.com/dr3y/dnaplotlib.git@master#egg=dnaplotlib
+bokeh
+matplotlib
+networkx
+fa2
+lxml
```
zoltuz: and this one?

dr3y: i think i left in only the essentials

dr3y: the binder is loading...

dr3y: ok it works

Issue #173, opened 2020-08-25
Depricate CRNlab


WilliamIX:
We should put in a deprecated warning so this functionality can be removed in a future release.

WilliamIX:
I just added warnings to CRNlab which say depricated and will be in my next PR

ayush9pandey:
Reminder to put different changes in different PRs for better tractability.

zoltuz:
This issue was solved in #177

Issue #172, opened 2020-08-12
Degredation_mRNA_MM global mechanism has recursive_species_filtering permanently set to True


Chryron:
The `Degredation_mRNA_MM` global mechanism does not have a recursive_species_filtering argument so by default it degrades any complexes containing RNA. 

```python
class Degredation_mRNA_MM(GlobalMechanism, MichaelisMenten):
    """Michaelis Menten mRNA Degredation by Endonucleases
       mRNA + Endo <--> mRNA:Endo --> Endo
       All species of type "rna" are degraded by this mechanisms, including those inside of a ComplexSpecies.
       ComplexSpecies are seperated by this process, including embedded ComplexSpecies. 
       OrderedPolymerSpecies are ignored.
    """
    def __init__(self, nuclease, name="rna_degredation_mm", mechanism_type = "rna_degredation", **keywords):
        if isinstance(nuclease, Species):
            self.nuclease = nuclease
        else:
            raise ValueError("'nuclease' must be a Species.")
        MichaelisMenten.__init__(self=self, name=name, mechanism_type = mechanism_type)

        GlobalMechanism.__init__(self, name = name, mechanism_type = mechanism_type, default_on = False,
                                 filter_dict = {"rna":True, "notdegradable":False}, recursive_species_filtering = True)
```

WilliamIX:
I have passed through these keywords (in Degredation_mRNA_MMand in Deg_Tagged_Degredation) - they now default to the right values but can be overwritten. This will be included in the Bug PR.

zoltuz:
Hi @Chryron , thanks for reporting this. This issue has been solved in #177  (thanks @WilliamIX ) and the dev branch now contains the bugfix.

If you want to give a try, you can install biocrnpyler using the code from the dev branch:
https://github.com/BuildACell/BioCRNPyler/wiki/How-to-install-BioCRNPyler#installing-the-dev-branch

PR #177, opened 2020-08-31
Bug fixes Part 1: Global Mechanisms and Depricate CRN lab

WilliamIX:
This PR fixes issue #173 by adding depricated warnings in most CRNlab functions.

This PR fixes issue #172 by passing through these keywords via the GlobalMechanisms constructors. Unittest were also added.

### DIFF
--- a/biocrnpyler/crnlab.py
+++ b/biocrnpyler/crnlab.py
@@ -23,12 +23,18 @@ class CRNLab(object):
         self.crn = None
         self.warning_print = True
 
+        warnings.warn("CRNLab is depricated and will cease to function with future releases.")
+
+
     def mixture(self, name, **kwargs):
         """
         Create a Mixture of a given name into the CRNLab object
         Specify the extract and buffer optionally
         Specify extra parameters to be loaded as dictionaries optionally
         """
+
+        warnings.warn("CRNLab is depricated and will cease to function with future releases: to produce a Mixture please use that Mixture's class constructor.")
+
         extract = kwargs.get('extract') 
         if 'warning_print' in kwargs:
             self.warning_print = kwargs['warning_print']
@@ -67,6 +73,9 @@ class CRNLab(object):
     def add_dna(self, dna = None, name = "", promoter = "", 
                 rbs = "", protein = "", initial_conc ="", 
                 final_conc = "", volume = 0, **kwargs):
+
+        warnings.warn("CRNLab is depricated and will cease to function with future releases: to add DNA to a mixture, please use Mixture.add_component.")
+
         if volume:
             self.volume += volume
         if dna:
@@ -86,6 +95,8 @@ class CRNLab(object):
         return self.Mixture
 
     def add_component(self, component):
+        warnings.warn("CRNLab is depricated and will cease to function with future releases: to add a Component to a mixture, please use Mixture.add_component.")
+
         self.Mixture.add_components(component)
         return self.Mixture
     def set_volumes(self):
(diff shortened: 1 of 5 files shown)


biocrnpyler/crnlab.py
```diff
@@ -23,12 +23,18 @@ def __init__(self, name = '', **kwargs):
         self.crn = None
         self.warning_print = True
 
+        warnings.warn("CRNLab is depricated and will cease to function with future releases.")
```
ayush9pandey, line 26: Typo here (and other places): It should be "deprecated".

biocrnpyler/global_mechanism.py
```diff
@@ -200,15 +200,20 @@ class Degredation_mRNA_MM(GlobalMechanism, MichaelisMenten):
        ComplexSpecies are seperated by this process, including embedded ComplexSpecies. 
        OrderedPolymerSpecies are ignored.
     """
-    def __init__(self, nuclease, name="rna_degredation_mm", mechanism_type = "rna_degredation", **keywords):
+    def __init__(self, nuclease, name="rna_degredation_mm", mechanism_type = "rna_degredation", 
+        default_on = False, recursive_species_filtering = True, filter_dict = None, **keywords):
```
ayush9pandey, line 204: Can we do this by passing **keywords throughout and using particular keyword arguments wherever they are needed (instead of having all keywords in the init call? 

In other words, if all of the keyword arguments make sense in this class's init call, then this is fine.

biocrnpyler/crnlab.py
WilliamIX: Will fix!

biocrnpyler/global_mechanism.py
WilliamIX: I thought of that, but decided to make all the normal global mechanism keywords visible because I thought it would help inspire people to reuse these mechanisms in different ways.

WilliamIX:
The failed coverage (I think) is just due to new lines of code [deprecated warnings being added]. I have added some additional checks to try to keep the percentage similar.

@ayush9pandey  or @zoltuz if you could review this PR then squash and merge that would be great.

biocrnpyler/crnlab.py
```diff
@@ -23,12 +23,18 @@ def __init__(self, name = '', **kwargs):
         self.crn = None
         self.warning_print = True
 
+        warnings.warn("CRNLab is deprecated and will cease to function with future releases.")
```
zoltuz, line 26: Can you change the warning to this?
```python
warnings.warn('"CRNLab is deprecated and will cease to function with future releases.", DeprecationWarning)
```

biocrnpyler/crnlab.py
```diff
@@ -23,12 +23,18 @@ def __init__(self, name = '', **kwargs):
...
+
+
     def mixture(self, name, **kwargs):
         """
         Create a Mixture of a given name into the CRNLab object
         Specify the extract and buffer optionally
         Specify extra parameters to be loaded as dictionaries optionally
         """
+
+        warnings.warn("CRNLab is deprecated and will cease to function with future releases: to produce a Mixture please use that Mixture's class constructor.")
```
zoltuz, line 36: same here

zoltuz (commented):
Looks good to me! Just add the "DeprecationWarning" type to the warnings and crnlab  and it is good to go then

WilliamIX:
@zoltuz added "DeprecationWarning" to all the CRNlab warnings. Also merged Dev into this branch to make sure merging is easy and there are no conflicts.

zoltuz:
Great! Let's wait for the checks to finish and I'll merge this PR into the dev

Issue #174, opened 2020-08-27
Wrong parameter passed into bioscrape annotation. Check images


ayush9pandey:
Model in SBML:
![image](https://user-images.githubusercontent.com/8139195/91480783-4f189a80-e858-11ea-9c4f-c01383aebd89.png)
Model loaded in bioscrape:
![image](https://user-images.githubusercontent.com/8139195/91480791-517af480-e858-11ea-8d76-e480b2243ecb.png)

Thanks to @Chryron for pointing this out!

zoltuz:
Hi @ayush9pandey, thanks for reporting this. This issue has been solved in PR  #178 (thanks @WilliamIX ) and the dev branch now contains the bugfix.

If you want to give a try, you can install biocrnpyler using the code from the dev branch:
https://github.com/BuildACell/BioCRNPyler/wiki/How-to-install-BioCRNPyler#installing-the-dev-branch

Issue #171, opened 2020-08-12
Parameter values with "-" character raise ValueError


Chryron:
If any parameter values contain the "-" character (e.g "1e-4"), the error below is triggered 
```python
if re.search('[a-d-f-z]', new_parameter_value, re.I) \
        or re.search('(^[1-9]+/[1-9]+)|(^[1-9]+e-?[0-9]+)|(^.?[0-9])', new_parameter_value, re.I) is None:
    raise ValueError(f'No valid parameter value! Accepted formats: 1.00 or 1e4 or 2/5, we got {new_parameter_value}')
```
since the `[a-d-f-z]` regex expression catches "-". The fix is just to change `[a-d-f-z]` to `[a-df-z]`

WilliamIX:
This is correct. SBML does not allow variables (including parameters) to use "-". Please use different parameter names.

Chryron:
Ah so instead of having a parameter be 1e-6, its only acceptable to use 0.000001?

ayush9pandey:
I think Nazmus meant that for parameter values (and the negative powers)

WilliamIX:
@Chryron are you passing a parameter directly to a reaction with a string for the parameter value or is this through a parameter file or parameter dictionary?

Chryron:
Through a parameter file but I can try the other ways. I believe the error is raised whenever its a string.

zoltuz:
Hi @Chryron, thanks for reporting this. This issue has been solved in PR #178 (thanks @WilliamIX ) and the dev branch now contains the bugfix.

If you want to give a try, you can install biocrnpyler using the code from the dev branch:
https://github.com/BuildACell/BioCRNPyler/wiki/How-to-install-BioCRNPyler#installing-the-dev-branch

PR #178, opened 2020-08-31
Bug fixes pt2

WilliamIX:
Fixes issue #174 by removing the duplicate parameter from the SBML parameter dictionary and removing local parameters k_foward/k_reverse if they are both created in a non-reversible reaction. Adds new unit tests to find duplicate bioscrape annotations.

Fixes issue #171 as described in that thread by changing the regex expression. Adds new unit tests.

fixes issue #179 by adding GeneralPropensity.pretty_print and an associated unit test.

### DIFF
--- a/Tests/test_parameter.py
+++ b/Tests/test_parameter.py
@@ -29,6 +29,9 @@ class TestParameter(TestCase):
 
         self.assertTrue(Parameter(parameter_name="None", parameter_value="1e2").value == 100)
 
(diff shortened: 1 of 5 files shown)


ayush9pandey (approved):
Changes seem fine to me. I am assuming you already checked the bioscrape annotations that get written and how they are read in the XML..

zoltuz:
All looks good to me! I've checked on my computer as well, all test has passed.

zoltuz (approved):

PR #195, opened 2020-10-02
refactoring roadrunner interface

zoltuz:
This PR refactors the roadrunner interface.

This new interface allows simulating CRNs with roadrunner (libroadrunner). Two main operation mode (determined by return_roadrunner flag): getting back numerical results or getting back the roadrunner model object for further analysis.

* refactor function name: simulate_with_roadrunner to keep it consistent with bioscraper
* unit test adjusted to detect both cases: libroadrunner is installed or not and
* an example Jupyter notebook is added

### DIFF
--- a/Tests/test_simulation_packages_import.py
+++ b/Tests/test_simulation_packages_import.py
@@ -54,12 +54,12 @@ def test_bioscrape_import_simulate_via_sbml():
 
 def test_libroadrunner_import():
     from biocrnpyler import ChemicalReactionNetwork
+    CRN = ChemicalReactionNetwork(species=[], reactions=[])
     try:
-        import numpy as np
-
-        CRN = ChemicalReactionNetwork(species=[],reactions=[])
+        import roadrunner
+    except ModuleNotFoundError:
         with pytest.warns(None) as record:
-            sim_results = CRN.runsim_roadrunner(timepoints=np.linspace(0, 10, 100), filename=None)
+            sim_results = CRN.simulate_with_roadrunner(timepoints=list(range(0,10)))
 
         assert sim_results is None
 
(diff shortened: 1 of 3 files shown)


Tests/test_simulation_packages_import.py
```diff
@@ -54,20 +54,23 @@ def test_bioscrape_import_simulate_via_sbml():
...
         with pytest.warns(None) as record:
-            sim_results = CRN.runsim_roadrunner(timepoints=np.linspace(0, 10, 100), filename=None)
+            sim_results = CRN.simulate_with_roadrunner(timepoints=list(range(0,10)))
 
         assert sim_results is None
 
         # only one warning was triggered
         assert len(record) == 1
         # check the warning message
         assert str(record[0].message) == "libroadrunner was not found, please install libroadrunner"
```
ayush9pandey, line 69: I guess this warning print out is fine but I am wondering why the 6 line set up to do this and not just print out or throw a normal warning if ModuleNotFoundError?

ayush9pandey (approved):

ayush9pandey:
Looks good to me, we can merge this now.

Tests/test_simulation_packages_import.py
zoltuz: Good question! it's an older test which I'm using here, I'd write it differently today, it's a bit baroque indeed :)

zoltuz: I went back and had a good look at it and simplified the code a bit, added some comments as well

WilliamIX:
This code all looks good to me.

WilliamIX (approved):

WilliamIX:
This all looks good to me.

PR #202, opened 2020-12-15
Component enumerator

### LINKED ISSUE #170: Component level mechanisms
Currently we have `Species` which can be involved in `Mechanisms`. However, wouldn't it be nice to have `ComponentMechanisms` which can convert `Components` into other `Components`? 

Some examples include:

- Transcription/ Translation that makes RNA_constructs out of DNA_constructs
- Integrases that convert attB sites into attL sites
- DNA_constructs that can be cleaved into two DNA_constructs of different composition (possibly by integrases? splicing, etc)
- epigenetics?

Compilation is something like this:

1. `Component` definitions (including `DNA_construct`)
2. Run `ComponentMechanisms` on the `Components` in the `Mixture` (generates special `Reactions` and new `Components`)
   a. `ComponentMechanisms` are contained inside certain `Component`s
3. get generated `Component`s and original `Component`s and feed them into the `Mixture`
4. normal `update_species`/`update_reactions` generation

dr3y:
This PR substantially reworks DNA_construct and generalizes the idea of Component Enumeration.

## New Classes:
* `ComponentEnumerator` which is Subclassed into `LocalComponentEnumerator` and `GlobalComponentEnumerator`
* `ConstructExplorer(LocalComponentEnumerator)` is subclassed into `TxExplorer` and `TlExplorer`. `ConstructExplorer` is an abstract class which iterates over a `Construct` (an assembly of DNAparts) and calls a number of abstract functions which are fleshed out in `TxExplorer` and `TlExplorer.`
* `TxExplroer` looks at a list of DNA_parts in a DNA_construct and figures out what will be transcribed in order to produce RNA_constructs
* `TlExplorer` looks at a list of DNA_parts in an RNA_construct and figures out what will be translated in order to produce different protein products.

## New Compilation Scheme:
* `Mixture` now calls `enumerate_components()` which in turn calls `Component.enumerate_components()` one every `Component`. This allows Components to generate additional Components using Component Enumerators. Later, GlobalComponentEnumeration which looks at interactions between components can be easily inserted into this framework to model things like Integrases.

## `DNA_construct` and `RNA_construct`
* Two example Components with Enumeration are `DNA_construct` and `RNA_construct`, which by default include a `TxExplorer` and `TlExplorer,` respectively.
* When `.enumerate_components()` is called on these classes, the Tx/Tl Explorer class figures out transcription and translation based upon the list of DNA_parts in each Construct. The construct also has a method, `combinatorial_enumeration()`, which generates combinatorial binding states so things like multiple promoters and complex binding interactions can be modeled.

## Test Cases:
* unit tests for `DNA_construct` and `RNA_construct` were created. These also test `TxExplorer` and `TlExplorer` explicitly.
* Combinatorial tests for `DNA_construct` and `RNA_construct` were created. These ensure that these Components Compile in a variety of Mixtures using a variety of DNA parts and produce "reasonable" CRNs.

closes #170


biocrnpyler/dna_construct.py
```diff
@@ -468,17 +715,13 @@ def changed(self):
         self.reset_stored_data()
         self.name = self.make_name()
     def update_reactions(self,norna=False):
-        
+        return []
```
zoltuz, line 718: nothing gets executed after this line. Is this a bug?

biocrnpyler/dna_construct.py
```diff
@@ -505,15 +751,16 @@ def __init__(self,
                 attributes=None,
                 initial_conc=None,
                 copy_parts=True,
+                component_enumerators = (TxExplorer(),),
```
zoltuz, line 754: I'm a bit confused by this. Tuples are immutable, so no new elements can **not** be added to the tuple. Is there a reason for such a construct? 
```python
component_enumerators = [TxExplorer()]
```
Consider using a  list instead

biocrnpyler/dna_construct.py
```diff
@@ -528,199 +775,20 @@ def __init__(self,
...
-        return self.made_proteins
-    def get_rnas(self):
-        """return all RNAs made during the latest exploration"""
-        return self.all_rnas
-
-
 class RNA_construct(Construct,RNA):
-    def __init__(self,parts_list,name=None,made_by="unknown",**keywords):
+    def __init__(self,parts_list,name=None,made_by="unknown",\
+                component_enumerators = (TlExplorer(),),\
```
zoltuz, line 784: same as above

biocrnpyler/dna_part_rbs.py
```diff
@@ -51,28 +51,35 @@ def update_reactions(self):
...
                 out_component.transcript = dna[mypos]
             else:
                 out_component.transcript = dna
-            #my_rna.get_species()
+        elif(proteins is None):
+            return None
+        else:
+            my_rna = None
+            if(len(proteins)==1):
+                my_rna = list(proteins.keys())[0]
```
zoltuz, line 65:
```python
my_rna  = list(proteins)[0]
```
does the job

biocrnpyler/dna_part_rbs.py
```diff
@@ -51,28 +51,35 @@ def update_reactions(self):
...
+            else:
+                #if this happens, it means we are being called from DNA
+                if(self.parent.get_species().material_type is not "dna"):
```
zoltuz, line 68: what happens when  `self.parent.get_species()` returns a list?

biocrnpyler/dna_part_rbs.py
```diff
@@ -51,28 +51,35 @@ def update_reactions(self):
...
+                    raise ValueError("something went wrong")
```
zoltuz, line 69: Give a bit more context to this error message. It's rather cryptic

biocrnpyler/dna_part_rbs.py
```diff
@@ -51,28 +51,35 @@ def update_reactions(self):
...
+                protein_species = []
+                for CDS in my_proteins:
+                    protein_species += CDS.update_species()
+
+                out_component.protein = protein_species
+                if(mypos is not None):
+                    out_component.transcript = dna[mypos]
+                else:
+                    out_component.transcript = dna
+                #my_rna.get_species()
```
zoltuz, line 84: remove if not needed

zoltuz (changes requested):
Please have a look at these comments/suggetions

biocrnpyler/dna_construct.py
dr3y: if i do it that way then your check for mutable default arguments goes off and the code fails. I could also make it none and set it in the body of init (if none) but this makes it more obvious what the default value is

dr3y: I did this because we are not allowed to use mutable lists in the default arguments. if you want to add elements you would pass in a list with more elements

biocrnpyler/component_enumerator.py
```diff
@@ -0,0 +1,38 @@
...
+        warn(f"Default update_components called for ComponentEnumerator = {self.name}.")
+        return []
+    def __repr__(self):
+        return self.name
+class LocalComponentEnumerator(ComponentEnumerator):
+    """a component enumerator's job is to create new components in a process similar to mechanisms.
+    A local component enumerator only cares about the single component that is passed in"""
+    def __init__(self,name:str):
+        ComponentEnumerator.__init__(self,name=name)
+class GlobalComponentEnumerator(ComponentEnumerator):
```
WilliamIX, line 27: If Local and Global Component enumerators end up being the same thing, we can just have one class ComponentEnumerator, then, right?

biocrnpyler/component_enumerator.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self,name:str):
+        """a component enumerator's job is to create new components in a process similar to mechanisms.
+    A global component enumerator takes in every component that is in the mixture. This is for complex enumeration that
+    cares about other components"""
+        ComponentEnumerator.__init__(self,name=name)
+    def enumerate_components(self,component=None,other_components=None):
```
WilliamIX, line 33: This is an interesting perspective. A component Enumerator belongs to a Component, so I see why it might make sense to have the component that "owns" the enumerator passed separately from the other components. Is there a use-case where we need it to work this way, as opposed to just passing in a list of components?

biocrnpyler/dna_construct.py
```diff
@@ -20,13 +20,284 @@
...
+        self.made_rnas = {}
+        self.all_rnas = {}
+        self.made_proteins = {}
+        self.current_rbs = None
+        self.possible_rxns = possible_rxns
+        self.direction=direction
+        self.second_looping = False
+        self.possible_directions = possible_directions
+    def enumerate_components(self,component):
+        #TODO the component we are given better be a dna_construct or something that can be txtled
```
WilliamIX, line 52: Great point. I think it is fair to enforce that this only works on DNAconstructs!

biocrnpyler/dna_construct.py
```diff
@@ -20,13 +20,284 @@
...
+        """return all RNAs made during the latest exploration"""
+        return self.all_rnas
+
+class TxExplorer(TxTlExplorer_CE):
+    def __init__(self,name="TxExplorer",possible_rxns=("transcription",),\
+                        direction="forward",possible_directions=("forward","reverse"),debug=False):
+        TxTlExplorer_CE.__init__(self,name=name,possible_rxns=possible_rxns,\
+                        direction=direction,possible_directions=possible_directions,return_proteins=False)
+
+class TlExplorer(TxTlExplorer_CE):
```
WilliamIX, line 295: I'm a little confused what is going on here - what is TxTlExplorer_CE versus TxExplorer vs TlExplorer?

biocrnpyler/dna_construct.py
WilliamIX: I would put this in the constructor of DNA_Assembly if None.

biocrnpyler/mixture.py
```diff
@@ -411,12 +414,32 @@ def compile_crn(self) -> ChemicalReactionNetwork:
...
+                comps_to_enumerate = outstanding_components
+                outstanding_components = []
+        if(len(outstanding_components) > 0):
+            warn("Mixture was left with unenumerated components "+str(', '.join(outstanding_components)))
+
         #Append Species from each Component
-        for component in self.components:
-            self.add_species_to_crn(component.update_species(), component)
+        for component in new_comp:
+            self.add_species_to_crn(remove_bindloc(component.update_species()), component)
```
WilliamIX, line 439: Could this be placed in CRN.add_species maybe?

biocrnpyler/mixture.py
```diff
@@ -411,12 +414,32 @@ def compile_crn(self) -> ChemicalReactionNetwork:
         #add the extra species to the CRN
         self.add_species_to_crn(self.added_species, component = None)
 
+        #local component enumeration
+        new_comp = []
+        comp_i = 0
```
WilliamIX, line 419: is comp_i used?

biocrnpyler/mixture.py
```diff
@@ -411,12 +414,32 @@ def compile_crn(self) -> ChemicalReactionNetwork:
...
+                new_comp += endpt_components
+                outstanding_components += unenumerated_components
+            if(len(outstanding_components) == 0):
+                #this happens if we are done early
+                break
+            else:
+                comps_to_enumerate = outstanding_components
+                outstanding_components = []
+        if(len(outstanding_components) > 0):
+            warn("Mixture was left with unenumerated components "+str(', '.join(outstanding_components)))
```
WilliamIX, line 435: Maybe we can make a helper function in mixture, like Mixture.enumerate_components() which does all this logic so the compile loop is cleaner?

biocrnpyler/dna_construct.py
zoltuz: The usual way to do this is something like this:
```python
def some_function(some_other_argument,component_enumerators=None):
   if component_enumerators is None:
      component_enumerators=(TxExplorer(),),)
```
You cannot enforce the structure or content of a list as the default argument of a function. But you can hint at it with type hints. For example:
```python
def some_function(some_arguemnt,component_enumerators: List[TxTlExplorer_CE]=None):
```
This serves as a type hint for the users, what kind of list is needed as an input to the function.

biocrnpyler/component.py
```diff
@@ -285,7 +285,14 @@ def update_reactions(self) -> List[Species]:
...
         return reactions
-
+    def update_components(self) -> (List,List):
+        """this is for component enumeration. Usually you will return a list of components that are
+        copies of existing ones (first list) and new components (second list). For example,
+        A DNA_construct makes a list of copies of its parts as the first output, and a list of RNA_constructs
+        as the second output.
+        An RNA_construct will make a list of copies of its parts as the first output, and a list of Protein
+        components as its second output (if it makes any proteins)"""
+        return [self], []
```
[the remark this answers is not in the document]
dr3y: it has to return self or else it won't work. Currently ALL components must return something from update_components or else they aren't placed in the list which the mixture evaluates

biocrnpyler/construct_explorer.py
```diff
@@ -0,0 +1,261 @@
...
+    def terminate_loop(self):
+        for rbs in self.current_proteins:
+            self.all_proteins[rbs] = self.current_proteins[rbs]
+        self.current_proteins = {} #reset who is translating
+
+    def return_components(self,component):
+        returnable_proteins = []
+        for rbs in self.all_proteins:
+            proteins = [a[0] for a in self.all_proteins[rbs]]
+            #print("protein directions")
```
zoltuz, line 254: please remove unused lines

biocrnpyler/dna_construct.py
```diff
@@ -37,19 +34,28 @@ def __init__(self,
...
+                component_enumerators = None,
                 **keywords):
         """this represents a bunch of parts in a row.
         A parts list has [[part,direction],[part,direction],...]
         Each part must be an OrderedMonomer"""
-        #TODO get_part, dna_part search engine like get_component from mixture
+        if(component_enumerators is None):
+            component_enumerators = []
+        self.component_enumerators = component_enumerators
+        '''
```
zoltuz, line 45: please remove unused code

biocrnpyler/dna_construct.py
```diff
@@ -37,19 +34,28 @@ def __init__(self,
                 attributes=None,
                 initial_conc=None, 
                 copy_parts=True,
```
zoltuz, line 36: copy parts is an unused argument

biocrnpyler/dna_construct.py
```diff
@@ -468,32 +432,9 @@ def changed(self):
...
-            #doing update_reactions, in case any of these parts have memory
-            reactions+= part.update_reactions()
-        for rna in proteins:
-            if(not rna == self):
-                #this part makes sure we don't do an infinite loop if we are in fact an RNA_construct
-                reactions += rna.update_reactions()
-        return reactions
+
 
 class DNA_construct(Construct,DNA):
```
zoltuz, line 439: DNA construct subclasses DNA but never initializes it. Same goes for RNA construct with RNA

biocrnpyler/dna_construct.py
```diff
@@ -528,199 +473,25 @@ def __init__(self,
...
-        rnadict = {self.my_promoter:self}
-        return rnadict, proteins
+        self.promoter = promoter
+        if component_enumerators is None:
+            from .construct_explorer import TlExplorer
+            component_enumerators = [TlExplorer()]
+
+        Construct.__init__(self=self,parts_list=parts_list,circular=False,name=name,\
+                                component_enumerators = component_enumerators,**keywords)
     #def get_species(self):
```
zoltuz, line 495: please remove unused code

biocrnpyler/utils.py
```diff
@@ -24,18 +24,23 @@ def rev_dir(dir):
...
+        """go through every species on a list and remove any "bindloc" attributes. This is used
+        to convert monomers with a parent polymer into the correct species after combinatorial binding
+        in things like DNAassembly and RNAassembly."""
+        
         out_sp_list = []
         for specie in spec_list:
-            #go through the species and remove the "bindloc" attribute
-            #I don't care about the binding now that I am done generating species
+            #waited species case used in reactions
             if(type(specie)==WeightedSpecies):
```
zoltuz, line 34: please use `isinstance()` instead

biocrnpyler/dna_part_rbs.py
```diff
@@ -49,30 +49,19 @@ def update_reactions(self):
         if self.protein is not None:
             reactions += mech_tl.update_reactions(transcript = self.transcript, protein = self.protein, component = self, part_id = self.name)
         return reactions
-    def update_component(self,dna=None,rnas=None,proteins=None,mypos=None):
+    def update_component(self,internal_species=None,**keywords):
```
zoltuz, line 52: **keywords argument is unused, please remove it

biocrnpyler/dna_part_promoter.py
```diff
@@ -78,24 +82,18 @@ def update_reactions(self):
                         component = self, part_id = self.name, complex = None,
                         transcript = self.transcript, protein = self.get_protein_for_expression())
         return reactions
-    def update_component(self,dna,rnas,proteins,mypos = None):
+
+    def update_component(self,internal_species=None,**keywords):
         """returns a copy of this component, except with the proper fields updated"""
-        if(dna.material_type == "rna"):
-            #Promoters only work with DNA
+        from .dna_construct import DNA_construct, RNA_construct
```
zoltuz, line 88: please move the package import to the top of the file!

biocrnpyler/dna_part_rbs.py
```diff
@@ -49,30 +49,19 @@ def update_reactions(self):
...
-        if(proteins is None):
-            raise ValueError("cannot update rbs {} when proteins is None".format(self))
-        my_rna = None
-        if(len(proteins)==1):
-            my_rna = list(proteins.keys())[0]
-        else:
-            #if this happens, it means we are being called from DNA
-            if(self.parent.get_species().material_type is not "dna"):
-                raise ValueError("something went wrong")
+        from .dna_construct import RNA_construct,DNA_construct
```
zoltuz, line 54: please move the package import to the top of the file!

biocrnpyler/dna_part.py
```diff
@@ -56,13 +59,15 @@ def __init__(self,name,mechanisms=None,parameters=None,**keywords):
...
         return Species(self.name, material_type="dna")
+
     def __repr__(self):
         myname = self.name
         if(self.position is not None):
             myname+="_"+str(self.position)
         if(self.direction =="reverse"):
             myname += "_r"
         return myname
+    '''
```
zoltuz, line 70: please remove unused code

zoltuz (changes requested):
It's getting into shape now. However, there are some unused bits of code here and there, please remove them. Also, there are lots of unused package imports at the top of the files. Please remove the ones that are not in use anymore. 
Other comments below:

biocrnpyler/construct_explorer.py
```diff
@@ -0,0 +1,261 @@
...
+from .component_enumerator import ComponentEnumerator,LocalComponentEnumerator, GlobalComponentEnumerator
+from .species import (ComplexSpecies, OrderedMonomer, OrderedPolymer,
+                      OrderedPolymerSpecies)
+from .utils import all_comb, remove_bindloc, rev_dir
+import logging
+
+
+class ConstructExplorer(LocalComponentEnumerator):
+    def __init__(self, name, direction="forward",possible_directions=("forward","reverse")):
+        ComponentEnumerator.__init__(self=self,name=name)
```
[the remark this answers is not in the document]
WilliamIX: Should be LocalComponentEnumerator!

biocrnpyler/dna_part_rbs.py
WilliamIX: We believe this will be needed when we implement global component enumerations.

biocrnpyler/dna_part_promoter.py
WilliamIX: @zoltuz This is necessary to avoid circular import loops. (DNA_construct.py depends on dna_part_promoter.py and visa-versa). If you know a better way to solve this, let us know!

biocrnpyler/dna_part.py
WilliamIX: repr is useful for debugging and used in some example notebooks I think?

dr3y (commented):
I enacted some changes, left others (for example, importing inside functions to avoid circular dependency

biocrnpyler/component_enumerator.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self,name:str):
+        ComponentEnumerator.__init__(self,name=name)
+
+class GlobalComponentEnumerator(ComponentEnumerator):
+    def __init__(self,name:str):
+        """a component enumerator's job is to create new components in a process similar to mechanisms.
+    A global component enumerator takes in every component that is in the mixture. This is for complex enumeration that
+    cares about other components"""
+        ComponentEnumerator.__init__(self,name=name)
+    def enumerate_components(self,components = None, **keywords):
```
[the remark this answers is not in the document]
dr3y: but the argument is now called components instead of component

biocrnpyler/construct_explorer.py
```diff
@@ -0,0 +1,261 @@
...
+class TxExplorer(ConstructExplorer):
+    def __init__(self, name = "TxExplorer", direction="forward",possible_directions=("forward","reverse")):
+        ConstructExplorer.__init__(self, name, direction=direction,possible_directions=possible_directions)
+
+
+    def reset_enumerator(self):
+        self.all_rnas = {} #must clear this before filling it up again
+
+    def initialize_loop(self, direction):
+        self.current_rnas = {} #Stores RNA's being examined in loop promoter --> [(DNA_part, direction) list]
```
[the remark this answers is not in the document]
dr3y: is there a nicer way to do it than in a comment?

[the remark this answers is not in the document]
zoltuz: you can initialize an empty dictionary with structure:
```python
self.current_rnas = {'promoter': None, 'part': None}
```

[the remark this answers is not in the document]
WilliamIX: this dictionary uses the representation of the promoter, repr(prom) --> (prom, direction), not the word "promoter".

WilliamIX (approved):

Tests/test_DNAconstruct.py
```diff
@@ -2,207 +2,85 @@
...
+    x = DNA_construct([P,T],mechanisms = mechs,parameters=parameters)
+    y = x.enumerate_components()
+    assert(y[0] == x[0]) #correct promoter is returned
+    assert(x[0].transcript == y[1].get_species()) #promoter in the DNA_construct has the correct transcript species
+    assert(y[1].promoter == x[0]) #rna is linked back to the DNA_construct's promoter
+    assert(y[1].promoter == y[0]) #correct promoter is returned by enumerate_constructs
+
+    #"incoherent" transcription
+    x = DNA_construct([[P,"reverse"],T],mechanisms = mechs,parameters=parameters)
+    y = x.enumerate_components()
```
ayush9pandey, line 46: Can we have helpful comments in enumerate_components about what it returns? I think it is defined at multiple places but it wasn't really clear to me what this will return in different cases.

ayush9pandey: (also because this function is the one that I see has been used so many times in tests)

Tests/test_DNAconstruct.py
```diff
@@ -2,207 +2,85 @@
...
+    assert(y[1].promoter == y[0]) #correct promoter is returned by enumerate_constructs
+
+    x = DNA_construct([[P,"reverse"],[T,"reverse"],],mechanisms = mechs,parameters=parameters,circular=True)
+    y = x.enumerate_components()
+
+    z = DNA_construct([T,P],mechanisms = mechs,parameters=parameters,circular=True)
+    e = z.enumerate_components()
+    assert(y[1]==e[1]) #different DNA constructs lead to the same RNA construct
+    assert(y[0]==x[0]) #correct promoter is working
+    assert(e[0]==z[1]) #correct promoter is working
```
ayush9pandey, line 86: Please add a blank line at the end of the file (for every file).

Tests/test_RNAconstruct.py
```diff
@@ -0,0 +1,61 @@
...
+    #minimal RNA transcription
+    x = RNA_construct([[R,"reverse"],C],mechanisms = mechs,parameters=parameters)
+    y = x.enumerate_components()
+    assert(y==[])
+
+    #minimal RNA transcription
+    x = RNA_construct([R,[C,"reverse"]],mechanisms = mechs,parameters=parameters)
+    y = x.enumerate_components()
+    assert(y==[x[0]])
+    assert(y[0].transcript.parent==x.get_species())
```
ayush9pandey, line 61: All these tests make sense to me and are kinda a good way to introduce someone new to these features in BioCRNpyler.

Tests/test_DNAconstruct.py
dr3y: why? (i'll do it, just curious)

ayush9pandey: [Nothing serious.](https://stackoverflow.com/questions/56728554/got-a-warning-saying-pep-8no-newline-at-end-of-file-is-this-affecting-my-cod) It's a [PEP8 guideline ](https://www.python.org/dev/peps/pep-0008/) because of how some files are read by different programs. Also, Github points this out in the PR when you go to the "Files Changed" tab.

Issue #198, opened 2020-10-08
Pretty print bug for propensity


WilliamIX:
![image](https://user-images.githubusercontent.com/332650/95513498-4e067d00-096f-11eb-82f8-0518fa1ea165.png)

Rate String pretty_print with proportional hill positive is missing a "/K"

PR #203, opened 2020-12-15
Fixes printing bugs

WilliamIX:
The propensity format for both the rate formula and the pretty_print text has been fixed to match in all propensity cases.

The Building CRNs Direclty example notebook has also been fixed to reflect these changes.

Tests have been corrected to reflect these changes.

Fixes issue #198


zoltuz (approved):
Looks good to me!

Issue #176, opened 2020-08-27
Add logic into Combinatorial Mixture/Component Testing


WilliamIX:
For example: a DNA_assembly should have a reaction with DNA as an input and a Transcript as an output.

PR #204, opened 2020-12-15
Combinatorial CRN validation via Component_CRN_validation function

WilliamIX:
This function looks at Components based on their type and ensures that the CRN has reactions with reasonable input and output species. Used in combinatorial testing to ensure that not only do all core Components compile but they produce reasonable CRNs.

Resolves issue #176

### DIFF
--- a/Tests/test_mixtures_and_components_combinatorially.py
+++ b/Tests/test_mixtures_and_components_combinatorially.py
@@ -184,7 +184,7 @@ class CombinatorialComponentMixtureTest(TestCase):
 					assert validate_sbml(document) == 0
 
 					#Validate the CRN topology
-					Component_CRN_validation(CRN, C, M)
+					Component_CRN_validation(CRN = CRN, component = C, mixture = M)
 
 		except Exception as e:
 			error_txt = f"Instantiating Component {comp} in Mixture {mixture} with args {args}. \n Unexpected Error: {str(e)}."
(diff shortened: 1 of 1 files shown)


Tests/test_mixtures_and_components_combinatorially.py
```diff
@@ -136,7 +198,7 @@ def test_dna_assemblies_in_mixtures(self):
 				for (rbs, args) in self.rbs_classes:
 					R = rbs(**args)
 					for (mixture, args) in self.mixture_classes:
-						A = DNAassembly("assembly", promoter = P, rbs = R)
+						A = DNAassembly("G", promoter = P, rbs = R)
```
zoltuz, line 201: If you change the first argument to a keyword argument, then we could avoid using the mix of positional and keyword arguments in function calls. (Same suggestion applies to other calls of `DNAassembly` and `Component_CRN_validation`). 
```python
A = DNAassembly(name='G', promoter = P, rbs = R)
```
It would make the code more maintainable in the long run. (Please don't change other function calls, so we can keep the scope of the PR limited to the original issue.) Many Thanks!

zoltuz (approved):
Apart from one suggestion, it looks good to me!

Tests/test_mixtures_and_components_combinatorially.py
WilliamIX: Done. Will merge now.

Issue #15, opened 2018-10-27
Support for Linear vs Plasmid DNA mechanisms


WilliamIX:
Mechanisms to add: 
Linear DNA should be degraded
GamS degredation inhibition

Should this be done with global mechanisms? Some other way?

WilliamIX:
Note that this can be succinctly modeled in by using attributes and global mechanisms.

dr3y:
circular is an attribute that a `DNA_construct` has, and right now it only affects how we are detecting transcription and translation. `OrderedPolymers` can have attributes but making an `OrderedPolymer` have attribute circular does nothing. What it *should* do though, is make sure that equality between circular constructs works when they are rotated. Likewise, equality of reverse constructs should work.

WilliamIX:
This can just be done via an example notebook with existing mechanisms

WilliamIX:
Do not need to natively support this - can be implemented in 3 lines. Maybe add an example?

PR #206, opened 2020-12-16
Linear dna degredation example

WilliamIX:
This PR adds an example of GamS blocking linear DNA degradation to the Global Mechanisms examples, per issue #15


zoltuz:
Something is still not okay when all the visualization packages are installed: 
Screenshot 2020-12-16 at 16 48 17

WilliamIX:
Problem fixed.

PR #209, opened 2020-12-17
Initial concentrations

WilliamIX:
This PR changes how initial_condition_dict work.
1. CRNs now have an initial_condition_dict (property). This can be set via the constructor, at compilation time via Mixture.Compile_CRN, or via the method CRN.update_initial_condition().
2. Mixtures and Components now store initial conditions as parameters using the mechanism name "initial concentration", part_id = mixture.name, and parameter name = str(species) (or if a species comes from a component.get_species(), component.name).
3. Components still have an initial concentration property, however this just adds a parameter to their parameter_database [kept for backward compatability]

Implementing these changes has resulted in many tests cases being altered, new tests cases (in particular the test_initial_condition file) as well as minor to underlying code regarding parameter searching and sbml writing, and changes to some example notebooks.


biocrnpyler/chemical_reaction_network.py
```diff
@@ -29,11 +30,12 @@ class ChemicalReactionNetwork(object):
     k \Prod_{inputs i} (S_i)!/(S_i - a_i)!
     where a_i is the spectrometric coefficient of species i
     """
-    def __init__(self, species: List[Species], reactions: List[Reaction], show_warnings=False):
+    def __init__(self, species: List[Species], reactions: List[Reaction], initial_condition_dict = None, show_warnings=False):
```
zoltuz, line 33: if you add a type hint here, that would help us in the long run:
```python
initial_condition_dict: Dict[Species,numbers.Real]
```
you might need to import numbers package the top of the file.

biocrnpyler/chemical_reaction_network.py
```diff
@@ -137,16 +164,32 @@ def pretty_print(self, show_rates = True, show_material = True, show_attributes
...
+        species_sort_list = [(parameter_to_value(ics(s)), s) for s in self.species]
+        species_sort_list.sort()
+        species_sort_list.reverse()
+        for sind, (init_cond, s) in enumerate(species_sort_list):
+            init_cond = ics(s) 
+
+            txt += s.pretty_print(show_material = show_material, show_attributes = show_attributes, **kwargs)
+
+            if show_initial_condition:
+                txt += f" (@ {parameter_to_value(init_cond)}),  "
```
zoltuz, line 180: do you have a helper function for such pretty_print in the ParameterDatabase class already?
If not, it might be better to write a slightly more general code in ParameterDatabase and call it with initial condition parameterization. Then such code can be reused elsewhere

biocrnpyler/mixture.py
```diff
@@ -303,63 +303,55 @@ def get_parameter(self, mechanism, part_id, param_name):
...
+            if not isinstance(s, Species):
+                raise ValueError(f"{s} is not a Species! Can only find initial concentration of a Species.")
+
+            init_cond = None
+            #1 Check the component
+            if component is not None:
+                init_cond = component.get_parameter(param_name = str(s), part_id = self.name, mechanism = "initial concentration", check_mixture = False, return_none = True)
 
-        s.initial_concentration = init_conc
+                if init_cond is None and component.get_species() == s:
```
zoltuz, line 339: it's called init_cond here. Is it intentional?

biocrnpyler/chemical_reaction_network.py
```diff
@@ -77,6 +79,31 @@ def add_reactions(self, reactions: Union[Reaction,List[Reaction]], show_warnings
...
+            init_cond_dict = {}
+        elif isinstance(init_cond_dict, dict):
+            for s in init_cond_dict:
+                if s not in self.species:
+                    raise ValueError(f"Trying to set the initial condition of a Species {s} not in the CRN")
+
+        self._initial_condition_dict = init_cond_dict
+
+    
+    def update_initial_condition(self, init_cond_dict):
```
zoltuz, line 98: I'm a bit confused by this, there's already a setter/getter for the initial concentration dict, now we have an `update_initial_condition` function. Why do we need the two?

zoltuz (commented):
I'm a bit confused by the initial_concetartion vs initial_cond lingo. Why the two remained after the refactoring?

biocrnpyler/chemical_reaction_network.py
WilliamIX: The reason I use this parameter_to_value helper function is because the initial conditions can be parameters, or numbers, and this takes care of both cases.

biocrnpyler/chemical_reaction_network.py
WilliamIX: added Dict[Species, Union[numbers.Real, Parameter]].

biocrnpyler/chemical_reaction_network.py
WilliamIX: This is a standard issue I have with getters and setters for dictionaries. The setter overwrites the entire dictionary, which is good at instantiation but typically not what we want to occur. The update_initial_condition updates the dictionary without overwriting it, in other words a subset of keys can be added or changed and the other key, value pairs will be untouched.

biocrnpyler/mixture.py
WilliamIX: I called it initial_condition of a CRN and initial concentration inside the ParameterDatabase. I tried to be consistent in that regard to keep back compatibility with Component.initial_concentration.

In my head, CRNs have an initial condition because they can be deterministic or stochastic. However, I figure that calling parameters initial concentration will be more helpful to users.

zoltuz: I think having a consistent naming convention in the code helps with code maintenance. Even if certain names are not scientifically accurate.

biocrnpyler/chemical_reaction_network.py
zoltuz: You can implement a setter anyway you want.  I'd just call self._initial_condition_dict.update(new_dict) after basic checks.

biocrnpyler/chemical_reaction_network.py
zoltuz: Thanks!

biocrnpyler/chemical_reaction_network.py
zoltuz: I see, in that case, my comment was not clear. Can you hide `more lines of code` inside that helper function?

biocrnpyler/mixture.py
WilliamIX: Ok, I will change to initial_concentration everywhere

biocrnpyler/chemical_reaction_network.py
WilliamIX: Don't think so. Parameters don't know what they are used for, and in some cases the init_cond is actually just a number. I don't want to return a string, because I use that helper function elsewhere when dealing with parameters which I want to treat as numbers.

biocrnpyler/chemical_reaction_network.py
WilliamIX: But then it becomes impossible to completely overwrite the dictionary.

WilliamIX: Ok, will change it. To reset the dictionary to emtpy you now have to do CRN.initial_concentration_dict = None

zoltuz:
Two comments:
- There's an unused `initial_condition_vector()` function in chemical_reaction_network.py line 191. Please remove it. (There's also an associate test: `test_initial_condition_vector()` in test_chemical_reaction_network.py)
- Could you please rename `initial_condition_dict` in `simulate_with_roadrunner()`? Thanks! I wrote the function before the remaining PR, hence the old naming convention.

Just an idea: maybe we could call the initial concentration variable simply` initial_concentration` without the dict. Probably, we don't need to encode the data structure in the variable name. What do you think? (we gain 5 characters! :))

zoltuz (approved):
I left a separate comment with some suggestions. These changes look good.

PR #211, opened 2020-12-18
Compartments feature 

### LINKED ISSUE #200: Compartments
Species should be able to have an SBML compartment tag. This won't do anything when simulated with bioscrape, but will work with other SBML simulators and SubSBML.

@ayush9pandey can you help spec out what this looks like? I have a couple of thoughts....

1. Compartments can just become attributes and be automatically incorporated into Species' names. Although this isn't what SBML intended, it will make everything work with bioscrape.
2. In deterministic CRNs, the volume of a compartment doesn't matter....so for now we can ignore that. If compartments are used with the stochastic=True case, we should raise a warning.
3. When writing SBML, biocrnpyler can look over all the species and make all the compartments automatically.

In the future, it would be cool for Bioscrape to include compartments, but this would be a fairly serious new feature.

ayush9pandey:
With this PR, BioCRNpyler can be used to specify species compartments as discussed in #200.

The basic usage is simple:
```
S1 = Species("S1", compartment = "new_compartment")
```

This command creates a species S1 and internally creates a new Compartment object with a default volume and dimension for this species.  A more advanced usage would be: 
```
new_compartment = Compartment("new_compartment", volume = 2)
S1 = Species("S1", compartment = new_compartment)
```

I have added a new jupyter notebook example that demonstrates this feature by creating a membrane model to model the transport of IPTG across two compartments. 

closes #200


biocrnpyler/species.py
```diff
@@ -597,18 +624,18 @@ class OrderedComplexSpecies(ComplexSpecies):
...
         if len(Species.flatten_list(species)) <= 1:
             raise ValueError("chemical_reaction_network.complex requires 2 "
                              "or more species in its constructor.")
         self.species = species
 
         #Call the Species superclass constructor
-        Species.__init__(self, name = name, material_type = material_type, attributes = attributes, initial_concentration = initial_concentration)
-
+        Species.__init__(self, name = name, material_type = material_type, attributes = attributes, 
+                        compartment = compartment, initial_concentration = initial_concentration)
```
zoltuz, line 637: tab?

zoltuz (commented):
Just some minor issues. Please have look.

zoltuz:
I guess the network plotting feature should also show the compartment's name when it's not the same for the whole network? Otherwise, it is misleading.

Tests/test_compartments.py
```diff
@@ -0,0 +1,21 @@
+from unittest import TestCase
+from biocrnpyler import Species, Compartment
+import pytest
+
+
+class TestSpecies(TestCase):
+
+    def test_compartment_initialization(self):
```
zoltuz, line 8: Probably, it's the best not to mix unittest and pytest packages. You can just write a pytest:

```python
def test_compartment_initialization():
    # tests naming convention repr without species type or attributes
      s_name = 'test_species'
      c_name = ''test_compartment''
    species = Species(name=s_name)
    compartment = Compartment(name=c_name)
    species.compartment = compartment
    assert compartment.name == c_name
    assert species.compartment.name == compartment.name
```

Tests/test_compartments.py
```diff
@@ -0,0 +1,21 @@
...
+        species.compartment = compartment
+        self.assertEqual(compartment.name, 'test_compartment')
+        self.assertEqual(species.compartment.name, compartment.name)
+
+        # tests naming convention for species with name and compartment
+        compartment = Compartment(name='test_compartment', spatial_dimensions = 1, volume = 1e-4)
+        self.assertEqual(compartment.spatial_dimensions, 1)
+        self.assertEqual(compartment.volume, 1e-4)
+        with pytest.raises(ValueError):
+            compartment = Compartment(name='2test_compartment')
```
zoltuz, line 21:
```python
with pytest.raises(ValueError):
            Compartment(name='2test_compartment')
```
return variable is not needed

biocrnpyler/component.py
```diff
@@ -291,7 +291,7 @@ def get_initial_condition(self, s):
 
         1. mixture.name, repr(s) in self.initial_condition_dictionary
         2. repr(s) in self.initial_condition_dictionary
-        3. If s == self.get_species and self.initial_con is not None: self.initial_conc
+        3. If s == self.get_species and self.initial_con is not None: self.initial_concentration
```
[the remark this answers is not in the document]
WilliamIX: Yes - we need to make sure that the initial_concentrations PR takes precedent. Maybe we can merge dev into this one first?

biocrnpyler/compartments.py
```diff
@@ -0,0 +1,63 @@
...
+    def volume(self, volume: float):
+        self._volume = volume
+    
+    def __eq__(self, other):
+        """
+        Overrides the default implementation
+        Two compartments are equivalent if they have the same name, spatial dimension, and volume
+        :param other: Compartment instance
+        :return: boolean
+        """
```
WilliamIX, line 56: I am worried about how this will propogate through BioCRNpyler if we start using the feature. If two Species (or any subclass thereof) have different Compartments, they must be different right? It seems challenging to have to keep track of the compartment by these 3 variables. Maybe we require that each compartment have a unique name to simplify things?

biocrnpyler/species.py
```diff
@@ -222,6 +243,7 @@ def __eq__(self, other):
                             and self.name == other.name \
                             and set(self.attributes) == set(other.attributes)\
                             and self.parent == other.parent\
+                            and self.compartment == other.compartment\
                             and self.position == other.position:
```
WilliamIX, line 247: I want to confirm that this equality test is universal and will work with all Species [I'm worried that it might have been overwritten in something like OrderedPolymerSpecies. Just making sure we do the compartment check everywhere we need.

WilliamIX (changes requested):

zoltuz:
@ayush9pandey can you rebase/merge this branch with dev to see the changes from #209 ?

Tests/test_compartments.py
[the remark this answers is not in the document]
ayush9pandey: Done.

Tests/test_compartments.py
ayush9pandey: Could you explain why this is not a good idea and also why do we have the two different packages? I like the unittest framework better because it kinda seems that it stores the results nicely in its classes..but not sure if that's an advantage or how it could be an advantage...

biocrnpyler/compartments.py
ayush9pandey: Fixing this by only looking at compartment names when comparing equality. Also, throw an error if two compartments have same name but different volume or spatial dimension.

Tests/test_compartments.py
ayush9pandey: At other places in our testing, we have similar mixing of unittest and pytest.

zoltuz: In the second unittest case, you use pytest's check for exceptions instead of the unittest built-in one.  Hence, the mixing. It's a minor thing, you can keep it this way. 
That's the unittest's way: 
https://ongspxm.gitlab.io/blog/2016/11/assertraises-testing-for-errors-in-unittest/

ayush9pandey: Yeah, it was using pytest way earlier and I moved some of those functions to be the unittest way. Do we have a preference? In the long run, it might be good to decide on one thing that we slowly migrate towards.

zoltuz: My hidden agenda is to migrate to pytest eventually, e.g. I write all my new tests in pytest. But feel free to use either for now.

Tests/test_sbml.py
```diff
@@ -1,146 +1,172 @@
...
-def test_create_sbml_model():
-    #tests creating an SBML model
-
-    document, model = create_sbml_model()
+        document, model = create_sbml_model()
 
 
-def test_add_all_species():
-    #Tests add_species via add_all_species
+    def test_add_all_species(self):
```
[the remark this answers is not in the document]
ayush9pandey: It's a tab space.

biocrnpyler/species.py
ayush9pandey: Done.

Tests/test_compartments.py
ayush9pandey: OK then I will move to pytest. Although with unittest there's the `setUp` function that you can create inside the class that's helpful at times (not sure if it's possible with pytest as well).

Tests/test_sbml.py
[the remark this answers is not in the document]
zoltuz: oh, that's bad! https://www.youtube.com/watch?v=cUFJ1yVhJ6g&ab_channel=yopugiscool
Remove it pls.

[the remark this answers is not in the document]
ayush9pandey: I thought VScode automatically took care of things like this :D

biocrnpyler/species.py
ayush9pandey: I double checked and I think this change is made everywhere in species.py. If you find some place that I might have missed please let me know.

ayush9pandey:
> 
> 
> I guess the network plotting feature should also show the compartment's name when it's not the same for the whole network? Otherwise, it can be misleading for the user.

We don't have a way to automatically make those compartment circles yet in the plotting feature if I am not mistaken. Also, species str will have compartment name appended (if it's not the default compartment), so those names will give away the compartment while plotting too I suppose.

ayush9pandey:
I think most comments are resolved now. Also, the membrane model SBML generated by the membrane model ipynb in the Special Components folder looks correct to me.

I can add a few more tests before merging if needed. Please let me know if there are further changes.

Tests/test_sbml.py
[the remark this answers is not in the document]
zoltuz: me too. Anyhow, check your settings pls.

zoltuz:
> > I guess the network plotting feature should also show the compartment's name when it's not the same for the whole network? Otherwise, it can be misleading for the user.
> 
> We don't have a way to automatically make those compartment circles yet in the plotting feature if I am not mistaken. Also, species str will have compartment name appended (if it's not the default compartment), so those names will give away the compartment while plotting too I suppose.

I was thinking of just adding the compartment's name next to the Species on the network graph for now, e.g.
A[in Compartment 2] or similar.

biocrnpyler/propensities.py
```diff
@@ -219,7 +219,8 @@ def _translate_propensity_dict_to_sbml(self, model, ratelaw):
             propensity_dict_in_sbml['parameters'][param_name] = parameter_in_sbml.getId()
 
         for species_name, species in propensity_dict_in_sbml['species'].items():
-            propensity_dict_in_sbml['species'][species_name] = getSpeciesByName(model, str(species)).getId()
+            # propensity_dict_in_sbml['species'][species_name] = getSpeciesByName(model, str(species)).getId()
```
zoltuz, line 222: please, remove this unused code

biocrnpyler/propensities.py
```diff
@@ -347,7 +348,8 @@ def create_kinetic_law(self, model, sbml_reaction, stochastic, reverse_reaction=
         if not reverse_reaction:
             reactant_species = {}
             for w_species in crn_reaction.inputs:
-                species_id = getSpeciesByName(model, str(w_species.species)).getId()
+                # species_id = getSpeciesByName(model, str(w_species.species)).getId()
```
zoltuz, line 351: please, remove this unused code

biocrnpyler/propensities.py
```diff
@@ -356,7 +358,8 @@ def create_kinetic_law(self, model, sbml_reaction, stochastic, reverse_reaction=
         elif reverse_reaction:
             reactant_species = {}
             for w_species in crn_reaction.outputs:
-                species_id = getSpeciesByName(model, str(w_species.species)).getId()
+                # species_id = getSpeciesByName(model, str(w_species.species)).getId()
```
zoltuz, line 361: please, remove this unused code

biocrnpyler/sbmlutil.py
```diff
@@ -245,15 +245,16 @@ def add_reaction(model, crn_reaction, reaction_id: str, stochastic: bool=False,
...
+        species_id = str(input.species)
         reactant = sbml_reaction.createReactant()
         reactant.setSpecies(species_id)  
         reactant.setConstant(False)
         reactant.setStoichiometry(input.stoichiometry)
 
 def _create_products(product_list, sbml_reaction, model):
     for output in product_list:
-        species_id = getSpeciesByName(model, str(output.species)).getId()
+        # species_id = getSpeciesByName(model, output.species.name).getId()
```
zoltuz, line 256: please, remove this unused code

biocrnpyler/species.py
```diff
@@ -627,7 +634,7 @@ def __init__(self, species, name = None, material_type = "ordered_complex", attr
 
         #Call the Species superclass constructor
         Species.__init__(self, name = name, material_type = material_type, attributes = attributes, 
-                        compartment = compartment, initial_concentration = initial_concentration)
+                            compartment = compartment, initial_concentration = initial_concentration)
```
zoltuz, line 637: tab space

Tests/test_sbml.py
```diff
@@ -26,17 +25,12 @@ def test_add_all_species(self):
...
-<<<<<<< HEAD
         species = [S1, S2, S3, S4, C1, C2]
-        add_all_species(model, species)
-=======
-    species = [S1, S2, S3, S4, C1, C2]
 
-    #initial conditions
-    init_cond_dict = {S1:1, S2:2, S3:3, S4:4}
+        #initial conditions
+        init_cond_dict = {S1:1, S2:2, S3:3, S4:4}
```
zoltuz, line 31: tab space

zoltuz (commented):
just some minor issues

biocrnpyler/species.py
ayush9pandey: I used a autopep8 extension on this file `species.py` to fix this and other similar possible issues.. Let me know if it looks fine.

ayush9pandey:
> 
> 
> > > I guess the network plotting feature should also show the compartment's name when it's not the same for the whole network? Otherwise, it can be misleading for the user.
> > 
> > 
> > We don't have a way to automatically make those compartment circles yet in the plotting feature if I am not mistaken. Also, species str will have compartment name appended (if it's not the default compartment), so those names will give away the compartment while plotting too I suppose.
> 
> I was thinking of just adding the compartment's name next to the Species on the network graph for now, e.g.
> A[in Compartment 2] or similar.

I have added compartment printing in the species.pretty print line in plotting.py. @dr3y  could you take a look at the change in plotting.py to make sure that would do the trick.

ayush9pandey:
I've added some test code to resolve coverage issues with this PR. Hopefully all other issues are also fixed. Ready to merge from my end.

ayush9pandey:
If there aren't any more changes, can we merge this first so that #219 (that's built on top of this PR) can be reviewed seamlessly?

I have added as many tests as I could but the remaining negative diff coverage is due to uncovered lines that are out of my expertise.

biocrnpyler/species.py
```diff
@@ -82,8 +86,22 @@ def name(self, name: str):
...
-    
-    #Use OrderedMonomers getter
+    @property
+    def compartment(self):
+        return self._compartment
+
+    @compartment.setter
+    def compartment(self, compartment):
+        if compartment is None:
+            self._compartment = Compartment(name="default")
```
WilliamIX, line 96: Can this have name "" instead of "default"?

biocrnpyler/species.py
```diff
@@ -188,6 +214,9 @@ def pretty_print(self, show_material = True, show_attributes = True, **kwargs):
 
         txt += self.name
 
+        if self.compartment not in ["", None] and show_compartment:
+            txt += " in " + self.compartment.name + "."
```
WilliamIX, line 218: The compartment name "default" will always be printed in this case

biocrnpyler/species.py
```diff
@@ -153,15 +174,20 @@ def __repr__(self):
             for i in self.attributes:
                 if i is not None:
                     txt += "_" + str(i)
+        if self.compartment.name != 'default':
+            # Only add a compartment name if it is not the default one.
+            txt += "_" + self.compartment.name
```
WilliamIX, line 179: I would also make "" skipped

biocrnpyler/species.py
```diff
@@ -411,20 +450,24 @@ class ComplexSpecies(Species):
...
             raise ValueError("chemical_reaction_network.complex requires 2 "
                              "or more species in its constructor.")
-        self.species = species 
+        self.species = species
 
-        #call super class
-        Species.__init__(self, name = name, material_type = material_type, attributes = attributes)
+        # call super class
+        Species.__init__(self, name=name, material_type=material_type, attributes=attributes,
+                         compartment=compartment, initial_concentration=initial_concentration)
```
WilliamIX, line 470: Species no longer have initial concentrations.

biocrnpyler/species.py
```diff
@@ -543,9 +592,13 @@ def pretty_print(self, show_material = True, show_attributes = True, show_initia
             count = self.species.count(s)
             if count > 1:
                 txt += f"{count}x_"
-            txt += s.pretty_print(show_material = show_material, show_attributes = False)+":"
+            txt += s.pretty_print(show_material=show_material,
+                                  show_attributes=False)+":"
         txt = txt[:-1]
 
+        if self.compartment not in ["", None] and show_compartment:
```
WilliamIX, line 599: 'default'?

biocrnpyler/species.py
```diff
@@ -702,58 +771,64 @@ class OrderedPolymerSpecies(OrderedComplexSpecies,OrderedPolymer):
     sometimes it is convenient to pass an internal Species. Both will work from the point of view
     of any Mechanism.
     """
-    default_material="ordered_polymer"
-    def __init__(self,species, name=None, base_species = None, material_type = default_material, \
-                             attributes = None,circular = False):
+    default_material = "ordered_polymer"
+
+    def __init__(self, species, name=None, base_species=None, material_type=default_material,
+                 compartment=None, attributes=None, initial_concentration=0, circular=False):
```
WilliamIX, line 777: initial_concentration is coming back. Please double check this is removed everywhere.

biocrnpyler/compartments.py
```diff
@@ -0,0 +1,76 @@
...
+class Compartment():
+
+    """ A formal Compartment object for a Species in a CRN.
+     A Compartment must have a name. They may also have a spatial dimension (such as 2 for two-dimensional,
+     or 3 for three-dimensional) and the volume in litres. 
+     Note: The "default" keyword is reserved for BioCRNpyler allotting a default compartment. 
+     Users must choose a different string. 
+    """
+
+    def __init__(self, name: str, volume = 1e-6, spatial_dimensions = 3, **keywords):
```
WilliamIX, line 15: Will we need units here too, eventually?

WilliamIX (changes requested):
In general looks good. A couple of points:

If a user does not set explicitly compartments, no compartments should show up in BioCRNpyler or the SBML. The use of a "default" compartment named "default" seems counterintuitive. Can we just have None or "" for a name and have that be skipped, including in the SBML writing?

Also, this PR seems to put initial_concentrations back into a number of Species constructors. Please remove these.

ayush9pandey:
Having a compartment is mandatory in SBML. And we were already using the
"default" compartment name earlier as well before this PR. The only new
thing here is that it is set beforehand for Species whereas earlier all
species were given compartments during SBML writing.

ayush9pandey:
But yeah, we can change pretty print and other places so default is not
printed when not intuitive.

biocrnpyler/compartments.py
ayush9pandey: Added units to compartments. Thanks for the suggestion!

ayush9pandey:
All green finally! Thanks everyone for your comments. Merging this.

PR #212, opened 2020-12-18
Refactor enzyme class 

### LINKED ISSUE #148: merge Enzyme and MultiEnzyme classes
We could keep the business logic of ```MultiEnzyme``` and rename it to ```Enzyme```. I don't see a reason to keep both classes. 

Besides, the name is misleading ```MultiEnzyme``` handles multiple substrates and products with a **single** enzyme

zoltuz:
MultiEnzyme and Enzyme classes are merged because they had the same functionality. The signature of Enzyme has changed to substrates (plural) and products (plural), but it accepts a single string, Species or Component not just lists of those. 
 

- [X] Example notebooks are adjusted to the new signature.
- [X] Unit tests are implemented for the Enzyme class.

Closes #148


WilliamIX (approved):
Pending test completion, this looks good to me.

Issue #120, opened 2020-07-10
CRN Species Overhaul


WilliamIX:
@zoltuz is currently working on a CRN Reaction overhaul. At some point, I think we will need to do something similar for Species. 

## Design Question 1: Should CRN Species should be "univerisal" pointers ie if two Species are equivalent, they point to the same object?

Recapping a discussion with @zoltuz : the issue is that multiple Species objects (different python pointers) can be the same in terms of their defined `==` functionality due to their string representation. For example:

`S1, S2 = Species("S"), Species("S")` are two distinct objects which will be viewed as equal `S1 == S2` ==> `True`

However, if I now change S1, like `S1.material_type = "m"`, `S1 == S2` ==> `False`. Importantly, S1 is a different pointer than S2 and thereby does not effect S2. 

_In other words, does it make sense for changing S1 to also change S2 if S1 is created so that it is equal to S2? The use case here is that I compile a CRN and then change some species in that CRN. This change can then propagate back through the entire compiled CRN. In the current implementation, this will definitely not occur consistently because many equivalent species are constructed as different python objects._

## Design Question 2: How can we represent species with multiple binding sites (eg a long piece of DNA) to easily create bound Complexes where a binding Species binds to a specific Site? Does this require new representations of species beyond the string representation we currently use?

In particular, I am referencing a discussion in https://github.com/BuildACell/BioCRNPyler/pull/102 where Andrey and I discuss how to use Species to represent binding positions on a greater piece of DNA. I will reiterate the example the discussed in the PR as well as a crucial issue Andrey brought up that seems to imply we need an overhaul.

**Basic Idea:**
Introduce a new Species class SubSpecies(Species). SubSpecies can ONLY exist inside a ScaffoldSpecies (there is probably a better name) which is effectively an OrderedComplexSpecies made up of SubSpecies. Importantly, SubSpecies know about their containing ScaffoldSpecies and use it's name. However, when a SubSpecies is passed into ComplexSpecies or OrderedComplexSpecies, the resulting Species object is automatically placed inside of the ScaffoldSpecies which contains the Subspecies.

I will illustrate this with an example of two binding sites B1 and B2 in a polymer P which can be bound to A1 and A2, respectively. Note that this is a combinatorial system with 4 possible states.

`A1, A2 = Species("A2"), Species("A1")`

`B1, B2 = SubSpecies("B1"), SubSpecies("B2")`

[...]

PR #214, opened 2020-12-18
CRN Species and Reactions are "immutable" and test cases

WilliamIX:
- This PR makes CRN.species and CRN.reaction properties. 
- These properties can be set once. If a user tries to set them again, there is an error. Instead, individual species and reactions must be added via the CRN.add_species(..) and CRN.add_reactions(...) function (which already existed). 
- The getters on these properties return deep copied lists, to ensure that internal species and reactions cannot be accessed. 
- Finally a number of tests were made.

This should resolve the unresolved parts of issue #120 which deals with the scope of CRN species and how to ensure they are "synchronized" by simply protecting them from user meddling.


zoltuz (commented):
It looks okay, but I'm a bit confused why separate add_species **and** setter are needed. Could not be merged two into just one setter? (same for reactions) 

We can take care of this later, just wanted to ask, in case you want to have another look at it.

WilliamIX:
> It looks okay, but I'm a bit confused why separate add_species **and** setter are needed. Could not be merged two into just one setter? (same for reactions)
> 
> We can take care of this later, just wanted to ask, in case you want to have another look at it.

The setter will set the List CRN.species. I do not not want people to have direct access to this list. So much so, that the getter returns a copy of it.

In order to add species (either individually, or in a list) the user instead has to use the add species function. This is analogous to CRN.species += List[Species]. However, because I am not giving access to CRN.species directly, the user has to go through this secondary function.

PR #215, opened 2020-12-18
refactor Promoter and RBS initialization in DNAassembly

### LINKED ISSUE #150: merge two promoter initialization cases
Screenshot 2020-08-01 at 18 58 44

zoltuz:
Promoter and RBS initialization logic was exposed to the DNAassembly. This has been moved inside the respective classes, and from_promoter() and from_rbs() class methods are introduced to handle the different cases.
- [X] **unit test**: New Unittests are included. 
- [X] **examples**: No example was modified

closes #150


WilliamIX (approved):
I think this is a great solution to that problem!

Issue #130, opened 2020-07-25
Units in biocrnpyler


ayush9pandey:
This is an enhancement issue that we should probably work on later. Although bioscrape does not support units and ignores all units. But most other SBML simulators work with units in certain ways. Regardless, the job of biocrnpyler is to create SBML models so we should probably be creating dimensionally correct models too. Towards that end, I think we could have parameter files accept units (if the user desires so)..and then set those units accordingly in our SBML model. Also, we could give options to set the default concentration units for the SBML file. Currently, we set the concentration for species with substance unit = mole, and volume unit = litre (so mole/litre for concentration). 

This issue is also motivated by feedback from some recent users who were wondering what the units are for everything in biocrnpyler.

WilliamIX:
This is a good issue to work on and goes hand-in-hand with some of the Parameter Defaulting schemes I've been thinking about (in particular, when can we "guess a ball-park rate" even if there is no parameter file). It would be ideal if we could re-use someone else's unit conversion library - do you know if libSBML has any support for units?

ayush9pandey:
Writing units in SBML is simple: we already do that a bit when creating SBML files: https://github.com/BuildACell/BioCRNPyler/blob/4c2ec4e6abf54fc9ddc783286b8538f07f882ae8/biocrnpyler/sbmlutil.py#L51

I can work on making sure we have units for all parameters and all initial conditions..

WilliamIX:
This looks like a great package! I think we could even tie it in as an optional dependency. Here is my thought:
1. Have a function inside of the Parameter Class which takes units. These can be read from the parameter files or passed into the Parameter constructor.
2. if Pint is installed, have the ParameterDatabase automatically convert all Parameters to the same units when it returns them. This will mean defining a default set of units (which can be changed, perhaps via a keyword). 
3. If Pint isn't installed or there are no units for a Parameter, everything will be assumed to be the in the default units (current functionality)

WilliamIX:
Add optional units to the parameter objects (and files) which propagates into SBML.

ayush9pandey:
Hard code SBML unit definitions for the following units:
nL, uL, mL, L. NM, uM, mM, M, hour, minute

zoltuz:
#219 solved this

PR #219, opened 2020-12-24
Units in BioCRNpyler

ayush9pandey:
This PR adds units compatibility for BioCRNpyler (addressing #130 ). Summary of how it works:

1. `units.py` has a list of functions that create unit definitions (as required by SBML). These functions are factored in a way that new functions can be easily added with no changes required at any other place in the code base. 
2. So, to add new units, the user simply adds their units to their parameter file in a new column, adds a function corresponding to that unit id string in `units.py` (this is a simple 3 line function that can be created by copying one of the existing function for units).
3. Parameter class has a new attribute "unit" that goes with every Parameter object (we can discuss what's the best way to do this). This is the part which is incomplete and I need some help so that we can get this done quickly.
4. The tsv file Perfect Parameter File has a new column "units", we will use this to test the functionality to make sure units are correctly read and written to SBML.

So, the current to-do is a minor one: successfully read the Perfect Parameter file with units so that all parameters that have been read have a `param.unit` attribute. 

(tests pending - will add later once overall functionality is finalized).


zoltuz:
Screenshot 2020-12-24 at 11 13 25

something is not square in `test_sbml.py` @ayush9pandey

ayush9pandey:
> 
> 
> Screenshot 2020-12-24 at 11 13 25
> 
> something is not square in `test_sbml.py` @ayush9pandey

Yeah that's the part that's the "to-do" => the unit attribute in parameter.py needs to be written correctly and I created this PR to ask William's help on it first. Probably best for you to review once this PR is out of the draft.

ayush9pandey:
Marking the PR ready for review. The functionality is there now as described earlier and it is working (Thanks @WilliamIX).

I will add tests now, and resolve any review you guys have.

WilliamIX:
Can you merge Comparments then rebase this PR to dev+Compartments so I can see the code changes independently of Compartments in order to do a review? Thanks!

ayush9pandey:
> 
> 
> Can you merge Comparments then rebase this PR to dev+Compartments so I can see the code changes independently of Compartments in order to do a review? Thanks!

Done.

Sorry for being slow on this. Should be good for review now. I am adding tests to get the coverage up.

ayush9pandey:
Tests have been added. Please let me know if there are any suggestions.

Tests/test_units.py
```diff
@@ -0,0 +1,44 @@
...
+                    + libsbml.OperationReturnValue_toString(value).strip() + '"'
+                raise SystemExit(err_msg)
+        else:
+            return
+
+class TestSpecies(TestCase):
+    def test_units_initialization(self):
+        # tests naming convention repr without species type or attributes
+        document, sbml_model = create_sbml_model()
+        list_of_supported_units = ["nL", "uL", "mL", "L",
```
WilliamIX, line 37: Not sure how hard this would be, but could we parse things like nL/s to nL_per_second and automatically generate some of these combinations from the primitives?

biocrnpyler/parameter.py
```diff
@@ -30,21 +30,24 @@
...
     # Then checks if (mixture.name, component.name) is in the self.custom_initial_condition_dictionary
     # Then checks if (component.name) is in the self.custom_initial_condition_dictionary
 
-    # Then checks if (mixture.name, repr(species) is in the parameter dictionary
+    # Then checks if (mixture.name, repr(species)) is in the parameter dictionary
     # Then checks if repr(species) is in the parameter dictionary
-    # Then checks if (mixture.name, component.name) is in the parameter dictionary
+    # Then checks if (mixture.name, component.name)) is in the parameter dictionary
     # Then checks if component.name is in the parameter dictionary
     # Then defaults to 0
```
WilliamIX, line 47: Probably my bad leaving these comments in Parameter, but can you just delete them all? this information is no longer valid and details are given in the Mixture class set_initial_condition function.

WilliamIX (commented):
Looks good - left two small comments.

Tests/test_units.py
ayush9pandey: Sure, we could do that..Only a matter of writing some string parsing code. But automatically writing the unit definitions could be tricky from the string letters such as "n" in "nL" and "m" in "mL/s"  (you need that for 10^-9  or 10^-6) etc or even it could get complicated like "nanomole_per_litre_per_second" or other stuff.

I think creating new unit definitions manually is very easy since you just have to copy a function in units.py and create your own by changing the exponents of different things and using your own string identifiers.

biocrnpyler/units.py
```diff
@@ -0,0 +1,164 @@
+import libsbml
```
zoltuz, line 1: please add a license header.

zoltuz (commented):
It looks good to me, but I'm a bit confused by these helper functions. There are lots of them and hard to maintain them.

Something like this would be a bit simpler:
```python
valid_units = {'litre_per_mole_per_hour': {unit_kin: [libsbml.UNIT_KIND_SECOND, libsbml.UNIT_KIND_LITRE, libsbml.UNIT_KIND_MOLE], 
        unit_exponents:[-1,1,-1], unit_scale:[0,0,0], unit_multipler:[3600,1,1]}}
```
and 
```python
def create_new_unit_definition(model, user_input):
    check_user_input(user_input)
    # create new unit

def check_user_input(user_input)
    user_input in valid_units
```

biocrnpyler/units.py
ayush9pandey: > 
> 
> It looks good to me, but I'm a bit confused by these helper functions. There are lots of them and hard to maintain them.
> 
> Something like this would be a bit simpler:
> 
> ```python
> valid_units = {'litre_per_mole_per_hour': {unit_kin: [libsbml.UNIT_KIND_SECOND, libsbml.UNIT_KIND_LITRE, libsbml.UNIT_KIND_MOLE], 
>         unit_exponents:[-1,1,-1], unit_scale:[0,0,0], unit_multipler:[3600,1,1]}}
> ```
> 
> and
> 
> ```python
> def create_new_unit_definition(model, user_input):
>     check_user_input(user_input)
>     # create new unit
> 
> def check_user_input(user_input):
>     user_input in valid_units
> ```

Yep, this is a good idea, will do something like that.

ayush9pandey:
I have made the requested changes. Please review and let me know if there are any more comments.

zoltuz (approved):
Thanks for all the changes, it looks great!

ayush9pandey:
All tests are passing now with positive diff coverage. Hopefully this gets merged before the new year 🤞

zoltuz:
> All tests are passing now with positive diff coverage. Hopefully this gets merged before the new year 🤞

Your wish has been granted :D You could've wished for an island in the Bahamas or PhD instead.

PR #220, opened 2020-12-31
Polymer conformation

WilliamIX:
This draft PR introduces a new Species class, `PolymerConformation(ComplexSpecies)`, which is used to store a set of `OrderedPolymerSpecies `and a set of `ComplexSpecies `representing binding between `OrderedMonomers `of the `OrderedPolymerSpecies`.

_Why is this necessary? Currently, BioCRNpyler does not have an easy way to represent Polymers which have been bound together. Some simple examples include hairpins or Looping where a single polymer binds to itself, which occur in RNA secondary structures as well as providing a mechanistic basis for many DNA regulatory elements, and things like integrases where multiple pieces of DNA can bind and recombine. The goal of this Species representation is that it will allow for such conformations to be represented and modeled unambiguously as elements of a CRN_

Lots of files have changed, but most of these changes are due to bugs that were found in the course of building this feature and new/changed test files. The key new pieces of code are all in Species.py. They are the new PolymerConformation class and an expansion of Complex(..) class which produces ComplexSpecies of different sorts.

Main Changes:
1.  `MonomerCollection` class is a superclass of OrderedPolymer and contains a list of OrderedMonomers without any additional informations. OrderedMonomer.parent (and therefore, Species.parent) must be a MonomerCollection.
2. `PolymerConformation` class is a subclass Species and MonomerCollection. It contains a list of OrderedPolymerSpecies and a list of ComplexSpecies representing binding between OrderedPolymerSpecies' monomers. Tests have been added for this classes functionality.
3. `Complex` class function has changed to produce PolymerConformations when needed. Test cases have been added.
4. `remove_bindloc` has been made recursive so that the top level of Species' parent-child relationship is inserted into reactions and/or a CRN.
5. These changes have resulted in a number of smaller changes propogating throughout the code and numerous test cases.


dr3y:
In general I like this. I think we would need to address how combinatorial enumeration treats these constructs.
When you say `construct.update_combinatorial_complexes(list(active_components))` you are relying on differentiating linear polymers from complexes of those polymers. Combinatorial Enumeration does this:

1. create all species assuming that only one component (in `active_components`) is active
2. create "mix and match" polymers where you take subunits from above and combine them with subunits created at different positions

For 2 to work you must be able to sift through a list of polymers and figure out which ones are versions of the same polymer but with stuff bound at different locations. After all, we are interested in all of the ways that things can be bound to the same polymer. This is how I am currently using the `base_species` member. Each polymer has a simple `Species` object inside which is basically the "name" of the polymer which has nothing bound at any location. This is different from the `name` member since it is shared between many different versions of the same polymer (but in that case the `name` of these different versions must be different or else compilation won't make sense). 

How does that work for a `PolymerConformation`? I think a regular `Species` type of thing will not be enough.

I want to get rid of it altogether by allowing a `OrderedMonomer` to have a flag which says if it is part of a polymer or not. That will allow a polymer full of `ComplexSpecies` to dig out the proper `Species` that make up the polymer itself. And, it will allow creation of new polymers by feeding them `ComplexSpecies`, and the result will make sense.

What happens if a `ComplexSpecies` containing two monomers is fed in like this? 

We can avoid this problem in many ways but at the end of the day it seems important to me to keep the physical configuration of the polymer (covalent connection of polymer subunits) stored somehow. Allowing complex configurations of polymers just highlights this problem, I think.

WilliamIX:
Thanks for the feedback @dr3y. Here are some thoughts:

With regards to Combinatorial Enumeration: I was hoping we could reuse the combinatorial stuff in each DNAConstruct. Basically, for each construct in the PolymerConformation, do the Combinatorial Enumeration keeping any bound complexes constant (this will require the ability to freeze a combinatorial site, which should be possible). Each set of combinations from each DNAConstruct could then be combined to get the full combinatorial set. One catch of this is that the IntegraseEnumerator will have to call the TxExplorer and TlExplorer itself - but I think that is probably inevitable. In other words, a DNAConstruct with integrase sites will overwrite its TxExplorer with an IntegraseExplorer which figures out integrations and also controls an internal TxExplorer.

With regards to OrderedMonomer flags: this won't work, as you describe, with PolymerConformation because in the case there are multiple Polymers, it won't be clear which one an OrderedMonomer belongs to. Here is an alternative thought: what if we use the OrderedMonomer.parent to keep track of this? The parent of a ComplexSpecies is the OrderedMonomer to which it is bound. The Parent of an OrderedMonomer is an OrderddPolymerSpecies. If an OrderedPolymerSpecies is in a PolymerConformation, it's parent will be the conformation. The parent of a  ComplexSpecies in a PolymerConformation bound to multiple OrderedMonomers will have the PolymerConformation as its parent. This has the nice feature that "remove_bind_loc" can now work recursively until it finds a Species with no parent! Conceptually, what I am thinking here is to add a flexible and arbitrary hierarchical description to these networks of OrderedMonomers.

biocrnpyler/species.py
```diff
@@ -878,19 +796,461 @@ def replace(self, position, part, direction=None):
...
+            #monomer = copy.copy(child_species[0]) #Copy the monomer
+            #monomer.remove()
+            #new_species.insert(insertlocs[0], monomer) #Insert the monomer back into the Species
+
+            # Create an OrderedcomplexSepcies or ComplexSpecies
+            child = copy.copy(child_species[0])
+            child_direction = child_species[0].direction
+            child.remove() #Remove the child species, the new_complex will replace it
+            species.remove(child_species[0])
+            species.insert(insertlocs[0], child) #place the new child in the list
```
dr3y, line 1190: took me a while to figure this out but you are doing this because you preserve the order in ordered complex. You can add a comment to explain that.

dr3y (approved):
i think it looks good to me! Took me a bit to understand what was happening but it makes sense.

biocrnpyler/polymer.py
```diff
@@ -5,14 +5,45 @@
...
+    """
+    def __init__(self, monomers):
+        self.monomers = monomers
+
+    @property
+    def monomers(self):
+        return self._monomers
+
+    @monomers.setter
+    def monomers(self, monomers):
```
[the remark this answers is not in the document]
WilliamIX: Added tests!

biocrnpyler/polymer.py
```diff
@@ -32,10 +63,10 @@ def __init__(self,parts,default_direction=None):
             position = len(polymer)-1
             direction = partdir
```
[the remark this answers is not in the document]
WilliamIX: I removed the "direction" variable but kept partdir [this is used when the list comes in with elements like [Monomer, direction].

biocrnpyler/polymer.py
```diff
@@ -165,40 +202,34 @@ def __init__(self,direction=None,position=None,parent=None):
...
     def parent(self):
         return self._parent
+
     @parent.setter
     def parent(self, parent):
-        if parent is None or isinstance(parent, OrderedPolymer):
+        if parent is None or isinstance(parent, MonomerCollection):
             self._parent = parent
         else:
             raise ValueError(f"parent must be an OrderedPolymer. Recieved {parent}")
```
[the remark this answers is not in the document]
WilliamIX: fixed.

biocrnpyler/polymer.py
```diff
@@ -165,40 +202,34 @@ def __init__(self,direction=None,position=None,parent=None):
...
         return self._position
+
     @position.setter
     def position(self, position):
         if self.parent is None and position is not None:
             raise ValueError("{} cannot have a position if it has no parent".format(self))
-        elif self.parent is not None and position is None:
-            raise ValueError("{} is part of a polymer with no position!".format(self))
         else:
             self._position = position
```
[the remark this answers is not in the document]
WilliamIX: I think typically setters don't return anything, so I will follow that convention. I think "insert" is the function that does both at once, right?

biocrnpyler/utils.py
```diff
@@ -25,27 +25,27 @@ def rev_dir(dir):
     reversedict = {"forward":"reverse","reverse":"forward"}
     return reversedict[dir]
 
+def recursive_parent(s):
+    #Recursively goes through Species and gets the top level parent
+    if hasattr(s, "parent") and s.parent is not None:
+        return recursive_parent(s.parent)
```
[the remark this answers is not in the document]
WilliamIX: python will automatically throw a Recursion error if that happens. In theory, one could make an PolymerConformation who's parent is a PolymerConformation.....but you'd have to do such shenanigans by hand - I don't think it will happen without forcing it to happen.

biocrnpyler/species.py
```diff
@@ -104,15 +96,25 @@ def compartment(self, compartment):
     # Use OrderedMonomers getter
     direction = property(OrderedMonomer.direction.__get__)
 
+    @property
+    def direction(self):
```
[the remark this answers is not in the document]
WilliamIX: Good spot - I removed the above one - the checking for the attribute was important in some cases where Species didn't get a direction set.

biocrnpyler/species.py
```diff
@@ -878,19 +796,461 @@ def replace(self, position, part, direction=None):
...
+
+        # Below cycle through species and see if one has a parent. If it does, that means the species is
+        # in an OrderedPolymerSpecies and the Complex should be formed around it.
+        for i, specie in enumerate(species):
+            if(hasattr(specie, "parent") and (specie.parent is not None)):
+                #This adds to a growing list of parents, which will be placed in an PolymerConformation
+                # It is very important to deepcopy here because the underlying OrderedPolymerSpecies or PolymerConformation will be modified.
+                parent_species.append(specie.parent)
+                bindlocs.append(specie.position)
+                insertlocs.append(i)
```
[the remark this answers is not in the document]
WilliamIX: insertlocs tells where in the list species the 'specie' with a parent came from. There may be multiple "specie"'s all with parents and I keep track of each ones index in the list species in order to preserve the order of the list later. This is something I added to prevent Compex from changing the order of lists (which could matters when making OrderedComplexSpecies).

biocrnpyler/species.py
WilliamIX: done.

biocrnpyler/species.py
```diff
@@ -263,6 +266,19 @@ def __hash__(self):
...
         return item in self.get_species()
 
+    def contains_species_monomer(self, s):
+        """Checks if the Species has a monomer (Species) inside of it, 
+        but without checking Species.parent, Species.position, or direction. In effect, a
+        less stringent version of __contains__."""
+        s_copy = copy.deepcopy(s)
+        s_copy.remove()
+        for ss in self.get_species(recursive = True):
+            ss_copy = copy.deepcopy(ss)
```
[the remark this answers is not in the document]
WilliamIX: good spot!

Tests/test_polymer_conformation.py
```diff
@@ -0,0 +1,207 @@
+#  Copyright (c) 2020, Build-A-Cell. All rights reserved.
```
ayush9pandey, line 1: I guess the license files should be 2021 now :D probably a mass find-and-replace needed.

Tests/test_specific_mechanisms.py
```diff
@@ -0,0 +1,119 @@
...
+A catch all of tests to ensure different mechanisms behave as desired
+"""
+
+from biocrnpyler import Species, Complex
+
+#Test Metabolite Mechanisms
+from biocrnpyler import OneStepPathway
+
+def test_OneStepPathway():
+    osp = OneStepPathway()
```
ayush9pandey, line 12: why the naming convention here is not with underscores like One_Step_Pathway like for other mechanisms? or the other way round for other mechanisms?

biocrnpyler/polymer.py
```diff
@@ -107,44 +143,44 @@ def direction_invert(self,dirname):
             return dirname
 
     def __len__(self):
```
ayush9pandey, line 145: an easy potential test line to increase coverage if needed.

biocrnpyler/polymer.py
```diff
@@ -107,44 +143,44 @@ def direction_invert(self,dirname):
...
-        return len(self._polymer)
+        return len(self.polymer)
```
ayush9pandey, line 146: an easy potential test line to increase coverage if needed.

ayush9pandey:
I made some minor comments. I got lost in the technical details of how the new classes are working so a documentation would definitely be helpful if someone new has to use these features. 

Other than that I am wondering if this PR changes in any way the basic usage of Species across the software?

Tests/test_polymer_conformation.py
WilliamIX: Lets do this on all of BioCRNpyler in the dev-->master PR that is no doubt coming soon.

WilliamIX:
> I made some minor comments. I got lost in the technical details of how the new classes are working so a documentation would definitely be helpful if someone new has to use these features.

I tried to provide descriptions in the Complex and PolymerConformation doc string, as well as in some of the test cases. Were these inadequate? Use examples (in an ipython notebook) will be coming in a later PR when I wrap this Species in a Component. Rushing this one because there are co-dependencies with stuff Andrey is working on.

> Other than that I am wondering if this PR changes in any way the basic usage of Species across the software?

No - this PR does not change how normal Species, ComplexSpecies, or even OrderedPolymerSpecies work. These are all additional features built on top of those.

Tests/test_specific_mechanisms.py
WilliamIX: I think we have not been very consistent with our naming convention, unfortunately.

biocrnpyler/polymer.py
WilliamIX: changed

Issue #225, opened 2021-02-06
Check rate laws when writing SBML for successful libSBML operation


ayush9pandey:

WilliamIX:
@ayush9pandey can you finish this next week sometime?

ayush9pandey:
#231 attempts to fix this

PR #231, opened 2021-02-17
Fix for SBML kinetic law formula #225

ayush9pandey:
* Added a check to Math AST node object that gets written to SBML file to ensure that the rate law formula is correct and returns a valid kinetic law.
* Added tests to test_sbml


ayush9pandey:
All tests are passing now.

WilliamIX (approved):

PR #222, opened 2021-01-09
Better plotting

dr3y:
you can display little dnaplotlib construct images in network diagrams

https://user-images.githubusercontent.com/3620565/104080719-9f2c4400-51de-11eb-8c1b-347f0ebeb9d9.mp4

PR #238, opened 2021-05-28
Speed improvements

dr3y:
Several improvements to CRN compilation and simulation speed.

1) constructs made by integration are checked against existing constructs so that they aren't remaking constructs that already have been made
2) constructs now generate a unique circular-permutation and direction-independent string which is used to compare constructs that are about to be created through integration.
3) you can make integrase rules which don't allow intermolecular reactions, or intramolecular reactions specifically
4) unique id for SBML model generation is made once and then assigned to every species/reaction instead of every time for each species/reaction

FYI: this is built on top of #222  so that one should be merged first :D


zoltuz:
Thanks for this PR! I've tried to have a look, but as you said the code changes are built on top of #222, so it's hard to decipher the logic for now. I was also wondering if the four changes that you've made could be broken up into different PRs (ideally four). Alternatively, you can add more hints in the description above (with code examples). 
Thanks, @dr3y !

dr3y:
ok yeah that is a good idea. I should break it up more. I'm getting better at using rebase so hopefully it won't be too hard now :)

zoltuz:
> ok yeah that is a good idea. I should break it up more. I'm getting better at using rebase so hopefully it won't be too hard now :)

Great! Let me know if I can help.

PR #239, opened 2021-06-02
pre-generate sbml ids for species and reactions

dr3y:
pre-generate all unique ids for species and reactions before iterating through all species and reactions to improve sbml model generation speed

### DIFF
--- a/Tests/test_sbml.py
+++ b/Tests/test_sbml.py
@@ -229,8 +229,8 @@ def test_generate_sbml_model():
     assert len(model.getListOfReactions()) == len(crn.reactions)
 
     document2, model2 = crn.generate_sbml_model(check_validity=False)
-    assert(model==model2) #the same model is made if you dont check validity
-    assert(document==document2) #the same model is made if you dont check validity
+    assert len(model2.getListOfSpecies()) == len(crn.species) #the same model is made if you dont check validity
+    assert len(model2.getListOfReactions()) == len(crn.reactions) #the same model is made if you dont check validity
     
     # reversible needs to be off!
     # assert not model.getListOfReactions()[0].isSetReversible()
--- a/Tests/test_sbml.py
+++ b/Tests/test_sbml.py
@@ -228,6 +228,10 @@ def test_generate_sbml_model():
     # all reactions from the CRN are accounted for
     assert len(model.getListOfReactions()) == len(crn.reactions)
 
+    document2, model2 = crn.generate_sbml_model(check_validity=False)
+    assert(model==model2) #the same model is made if you dont check validity
+    assert(document==document2) #the same model is made if you dont check validity
+    
     # reversible needs to be off!
     # assert not model.getListOfReactions()[0].isSetReversible()
 
@@ -354,7 +358,6 @@ def test_sbml_basics():
     S2 = Species("S2")
     compartment = add_compartment(model, S1.compartment)
     check(compartment, 'add compartment')
-    print(trans.getValidIdForName(str(S1)))
     sbml_species = add_species(
         model, compartment, str(S1.name),trans.getValidIdForName(str(S1)), initial_concentration=10)
     sbml_species2 = add_species(
--- a/Tests/test_sbml.py
+++ b/Tests/test_sbml.py
@@ -338,6 +338,14 @@ def test_sbml_basics():
     # generate an sbml model
     document, model = create_sbml_model()
 
+
+    elementlist = model.getSBMLDocument().getListOfAllElements()
+
+    all_ids = getAllIds(elementlist)
+
+    trans = SetIdFromNames(all_ids)
+
+
     # If the document/model creation failed the following two lines
     # will catch it:
     check(document, 'create SBMLDocument object')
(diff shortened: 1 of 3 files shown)


biocrnpyler/sbmlutil.py
```diff
@@ -90,10 +105,10 @@ def add_all_species(model, species: List, initial_condition_dictionary: dict, co
                 initial_condition_dictionary[s])
         else:
             initial_concentration = 0
-        add_species(model=model, compartment=compartment,
-                    species=s, initial_concentration=initial_concentration)
+        add_species(model=model, compartment=compartment, species_name=str(s.name), 
+                            species_id=s_id, initial_concentration=initial_concentration)
 
-def add_species(model, compartment, species, initial_concentration=None, **kwargs):
+def add_species(model, compartment, species_name, species_id, initial_concentration=None,  **kwargs):
```
zoltuz, line 111: I was wondering why this one is better than figuring out the `species_id` inside `add_species()`?

dr3y: because it has to check through all the ids in the model already. this way it only checks them all once instead of every time you add a species

Tests/test_sbml.py
```diff
@@ -346,10 +354,11 @@ def check(value, message):
     S2 = Species("S2")
     compartment = add_compartment(model, S1.compartment)
     check(compartment, 'add compartment')
+    print(trans.getValidIdForName(str(S1)))
```
zoltuz, line 357: please remove this line

biocrnpyler/chemical_reaction_network.py
```diff
@@ -280,8 +280,8 @@ def generate_sbml_model(self, stochastic_model=False, show_warnings = False, **k
         :param keywords: extra keywords pass onto create_sbml_model() and add_all_reactions()
         :return: tuple: (document,model) SBML objects
         """
-        ChemicalReactionNetwork.check_crn_validity(self._reactions, self._species, show_warnings=show_warnings)
-
+        if(check_validity):
+            ChemicalReactionNetwork.check_crn_validity(self._reactions, self._species, show_warnings=show_warnings)
```
zoltuz, line 284: can you add one empty line after this for readablity? Thanks!

zoltuz: Screenshot 2021-06-06 at 19 37 44

biocrnpyler/chemical_reaction_network.py
```diff
@@ -271,7 +271,7 @@ def replace_species(self, species: Species, new_species: Species):
 
         return ChemicalReactionNetwork(new_species_list, new_reaction_list)
 
-    def generate_sbml_model(self, stochastic_model=False, show_warnings = False, **keywords):
+    def generate_sbml_model(self, stochastic_model=False, show_warnings = False, check_validity = True,**keywords):
```
zoltuz, line 274: sorry for asking these questions, but why you need this check_validity flag? Is it for speed improvement again?

biocrnpyler/sbmlutil.py
```diff
@@ -236,18 +249,21 @@ def add_reaction(model, crn_reaction, reaction_id: str, stochastic: bool = False
                           sbml_reaction=sbml_reaction, model=model)
         _create_products(product_list=crn_reaction.inputs,
                          sbml_reaction=sbml_reaction, model=model)
-
+
```
zoltuz, line 252: please remove the tab character.

zoltuz (commented):
Only minor comments. Please have look! 
I was wondering that should we add the `check_validity` flag to `simulate_with_roadrunner()` as well?

biocrnpyler/chemical_reaction_network.py
dr3y: yes correct. "Check validity" is pretty slow so my work around is just not do it :). I am not 100% sure why it is needed but @WilliamIX can probably tell you more

biocrnpyler/sbmlutil.py
dr3y: this does not exist in my editor :P

biocrnpyler/chemical_reaction_network.py
dr3y: i added `check_validity` for `simulate_with_roadrunner()`

biocrnpyler/chemical_reaction_network.py
WilliamIX: check_validity calls an SBML function (from lib_sbml) which ensures that the final model is valid. By default this should definitely be on - it is particular important in ensuring that general propensity strings "make sense". For specific test cases, this could be False if the test is not testing SBML writing.

biocrnpyler/sbmlutil.py
zoltuz: got it!

biocrnpyler/chemical_reaction_network.py
zoltuz: Okay, I got it now.

zoltuz (approved):
All good. Thanks for the explanations!

ayush9pandey (approved):

PR #240, opened 2021-06-02
speed improvements in integrase enumeration

dr3y:
improve integrase enumeration speed by making the check for existing integrase-generated dna_constructs faster.

Each construct, when created, will make a unique string that will be the same if that construct is reverse, or circularly permuted. This is called the `directionless_hash`.

When you use integrases to create a new DNA construct, instead of comparing the dna parts inside and try all possible circular permutations, you just compare these hashes, which is much faster


biocrnpyler/integrase_enumerator.py
```diff
@@ -8,7 +9,7 @@
 import itertools as it
 from .utils import combine_dictionaries
 import copy
-
+import time
```
zoltuz, line 12: please remove unused package import

biocrnpyler/integrase_enumerator.py
```diff
@@ -259,6 +260,13 @@ def reaction_allowed(self,site1,site2):
         if(tuple([site1.site_type,site2.site_type]) in self.reactions):
             return True
         return False
+    def reactive_sites(self):
```
zoltuz, line 263: This can be simplified a bit:
```python
def reactive_sites(self):
    return list(set([list(reaction) for reaction in self.reactions]))

**Review by zoltuz — COMMENTED** [review]
I left some comments. It would be good to see the coverage and the test results. 

There are very few (I could only find one) tests that invoke the ```Integrase_Enumerator``` class, and as far I can see, its methods don't have individual tests, so it is hard to get clarity on the new features.  

I haven't looked at the hash generation itself.

File: biocrnpyler/integrase_enumerator.py (hunk 2 above)
Comment (dr3y) [reply biocrnpyler/integrase_enumerator.py:263]: i don't think so because i am starting with things like this `[(a,b),(c,d),(a,b)]` and I want to get `[a,b,c,d]`. Doing list comprehension on this nested tuple thing does not flatten the list.

Comment (zoltuz) [reply biocrnpyler/integrase_enumerator.py:262]: Oh, I see, I'll try to test out both version with the new tests you added to get a better understanding.

**Comment by zoltuz** [comment]
Here is the comparison between the run time before this PR and with this PR
Before:
122013361-02398c80-cdb6-11eb-8cf1-873e2bb3b679
After:
Screenshot 2021-06-15 at 13 02 54

**Comment by dr3y** [comment]
interesting i wonder why the speed gain is not that high. In my own notebooks I get a much higher speed improvement. Also you should check integrase examples cell 6 or 7 (whichever one does the different recursion depth crn compilations). That one should have a bigger speed benefit because you really see it when you do higher recursion depth

**Comment by zoltuz** [comment]
> interesting i wonder why the speed gain is not that high. In my own notebooks I get a much higher speed improvement. Also you should check integrase examples cell 6 or 7 (whichever one does the different recursion depth crn compilations). That one should have a bigger speed benefit because you really see it when you do higher recursion depth

It did this on my laptop with other things going on.  We will see it perform in Travis or other third party places.

File: biocrnpyler/dna_construct.py [hunk 3]
```diff
@@ -397,12 +403,142 @@ def enumerate_components(self,previously_enumerated=None):
...
+    def hashless_reverse(cls,construct):
+        """create a reverse construct but don't calculate its hash (because that would make an infinite loop)"""
+        rev_con = [a.get_orphan() for a in construct]
+        for rev_part,origpart in zip(rev_con,construct):
+            rev_part.direction = {"forward":"reverse","reverse":"forward"}[origpart.direction]
+        rev_con = rev_con[::-1]
+        rev_con = Construct(rev_con,make_dirless_hash=False,circular = construct.circular)
+        return rev_con
+    @classmethod
+    def rotation_free_hash(cls,construct,return_rotation=False):
```
zoltuz, line 430: `rotation_free_hash()` is always used with `return_rotation=True`. Can you remove the `False` branch?

biocrnpyler/dna_construct.py
```diff
@@ -397,12 +403,142 @@ def enumerate_components(self,previously_enumerated=None):
...
+        curname = str(orphan)
+        curdir = part.direction
+        return curname+curdir
+    @classmethod
+    def get_partlist_hash(cls,partlist):
+        """creates a string containing the name and direction of all parts in a list of parts (but not their position)"""
+        partlist_str = '_'.join([str(a[0])+str(b) for a,b in zip(partlist,range(len(partlist)))])
+        return partlist_str
+    @classmethod
+    def hashless_reverse(cls,construct):
```
zoltuz, line 421: `create_hashless_reverse()` would be a better function name, I think

biocrnpyler/dna_construct.py
```diff
@@ -397,12 +403,142 @@ def enumerate_components(self,previously_enumerated=None):
...
+                    best_partlist = test_partlist
+        while(len(best_partlist)<len(construct)):
+            nextpart = circular_next(best_partlist[-1][1],construct)
+            best_partlist += [(Construct.get_partstring(nextpart),nextpart)]
+        if(return_rotation):
+            return Construct.get_partlist_hash(best_partlist),best_partlist[0][1].position
+        else:
+            return Construct.get_partlist_hash(best_partlist)
+    @classmethod
+    def direction_rotation_free_hash(cls,construct,return_transform = False):
```
zoltuz, line 474: `direction_rotation_free_hash()` is always used with `return_transform=True`. Can you remove the `False` branch?

biocrnpyler/dna_construct.py
```diff
@@ -397,12 +403,142 @@ def enumerate_components(self,previously_enumerated=None):
...
+                return forward_hash,1,pos
+            else:
+                return forward_hash
+        else:
+            if(return_transform):
+                return reverse_hash,-1,posrev
+            else:
+                return reverse_hash
+    @classmethod
+    def linear_direction_free_hash(cls,construct,return_transform = False):
```
zoltuz, line 490: Same here.
`linear_direction_free_hash()` is always used with `return_transform=True`. Can you remove the False branch?

biocrnpyler/dna_construct.py
```diff
@@ -397,12 +403,142 @@ def enumerate_components(self,previously_enumerated=None):
...
+            if(return_transform):
+                return rethash,1,0
+            return rethash
+        elif(winner == "reverse"):
+            rethash = Construct.get_partlist_hash(test_reverse_partlist)
+            if(return_transform):
+                return rethash,-1,0
+            return rethash
+    @classmethod
+    def omnihash(cls,construct,return_transform=False):
```
zoltuz, line 522: This is the most important as this is the interface to the hashing part. Can you remove the `return_transform=True` branch of the code? Otherwise, we need to document the return type and create tests for it.

zoltuz (changes requested):
Almost there. Sorry for being a pain in the butt. 
During the hash creation, there are unused branches of the code (see below). I don't see a strong case keeping them unless there's a use case/example in the code and tests for them.  More importantly, these branches are changing the return type of the function, which will cause lots of headaches down the line if we keep them. Long story short, let's keep the code we use now.

zoltuz (approved):
Great! Thanks for the changes.

PR #241, opened 2021-06-02
Allow disallow specific integrase reaction types

dr3y:
maybe sometimes you want to make an integrase reaction that can only react within the same piece of DNA and not with other copies of that DNA or you want an integrase that can only delete and can't integrate.

Well, now you can specify that by writing something like this:
`bxb1_nodel_mechanism = IntegraseRule("Bxb1",allow_deletion=False)`

Other options include
`allow_deletion`
`allow_integration`
`allow_inversion`

and if you want to prevent reactions with a copy of the same DNA you have to create a part (any part) that has an attribute called `"no_inter"`. That allows you to have many dnas in a single reaction where some of them can react with copies of themselves and others can't. (That part was already done in #234 but I thought I'd mention it here anyway)

### DIFF
--- a/Tests/test_integrase_enumerator.py
+++ b/Tests/test_integrase_enumerator.py
@@ -68,7 +68,9 @@ def test_integrase_rule():
         bad_rule.generate_products(delete[0],delete[2]) #attP->attB reaction is not possible with a bad rule
     
     aL = IntegraseSite("attL","attL",integrase="Bxb1")
+    aL.direction= "forward"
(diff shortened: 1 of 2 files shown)


zoltuz:
It seems to me, files are missing from this PR, can you double-check it, @dr3y ?

dr3y:
oops i guess most of the changes in this PR are already in dev :D

zoltuz (approved):
not much to review. Looks good.

PR #251, opened 2021-07-01
fix bug in intermolecular reaction generation

dr3y:
This is fixing a weird bug i found when you have an integrase site that can also bind something that is not integrase (for example, a repressor).

Integrase sites turn into different sites (attB + attP -> attL + attR), but they need to remember that they were bound to things before and after. Thus:

1) When you do intermolecular reactions (another copy of the same piece of DNA reacts) you need to make sure there is a new polymer made which has a distinct parent, or else the polymer transformation doesn't know that there should be two inputs
2) integrase sites which turn into other integrase sites need a special way to know that their parents need to change specifically when it is an intermolecular reaction
3) dna constructs need to take their name into account when comparing to see if two constructs are equal (because i want to have two identical constructs that are not equal, in the case of doing this intermolecular reaction)

### DIFF
--- a/biocrnpyler/dna_construct.py
+++ b/biocrnpyler/dna_construct.py
@@ -553,7 +553,7 @@ class Construct(Component,OrderedPolymer):
     def __eq__(self,construct2):
         """equality means comparing the parts list in a way that is not too deep"""
         #TODO: make this be a python object comparison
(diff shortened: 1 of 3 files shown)


WilliamIX (approved):

WilliamIX:
@dr3y are you ready to merge this?

PR #252, opened 2021-07-14
CombinatorialComplex and CombinatorialConformation

WilliamIX:
This PR predominantly adds two new classes to represent combinatorial binding of individual molecules (Complex) and PolymerConformations (CombinatorialConformation).

* CombinatorialConformation is a class which represents and enumerates binding and unbinding to form a PolymerConformation (with just a single Polymer inside). In effect, this enumerates pathways to create different secondary structures.
* CombinatorialConformationPromoter is a CombinatorialConformation that can also work as a Promoter
* There are also some changes made to the plotting function (basically passing through a bunch of keywords) which I needed to create a figure in the paper.
* Fixed some issues with PolymerConformation and the Complex function which can up as I tested these new classes.
* Fixed some underlying issues with GlobalMechanisms in Mixtures and set_species in Component which came up with the lac operon example in figure 3 of the paper


WilliamIX:
@dr3y @ayush9pandey @zoltuz need some help here - for some reason the build is failing immediately, but I haven't changed anything to the testing environment etc. Any thoughts? Thanks!

setup.py
```diff
@@ -6,7 +6,7 @@
 
 setup(
     name='biocrnpyler',
-    version='1.1.0',
+    version='1.1.1',
```
ayush9pandey, line 9: why does this PR update the version? we can do this in dev-master PR

.github/codecov.yml
```diff
@@ -0,0 +1,11 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 75
+        threshold: .5
+    patch:
+      default:
+        target: 75
+        threshold: .5
```
dr3y, line 10: should this file be in gitignore?

Tests/Unit/test_complex.py
```diff
@@ -92,20 +117,31 @@ def test_complex_with_multiple_polymers():
...
-    #This occurs when a Complex is formed around a Monomer in a Polymer in a PolymerConformation which does not include any other Monomers with parents.
+    #These tests show how the order of binding can matter. 
+    #Best practices is to put everything into a PolymerConformation before doing Complex if PolymerConformations are being used.
+
     a = Species('A')
     b = Species('B')
+    s = Species("S")
     p = OrderedPolymerSpecies([a,b,a])
-    pc = Complex([p[0], p[1]]).parent #get a PolymerConformation
+    pc0 = PolymerConformation(polymer = p) #Put the polymer inside a conformation
```
dr3y, line 127: why doesn't `Complex` already do that automatically?

Tests/Unit/test_polymer_conformation.py
```diff
@@ -6,18 +6,42 @@
...
+    p = OrderedPolymerSpecies([Species("m1"), Species("m2"), Species("m3")])
+    pc = PolymerConformation(polymer = p)
+    c1 = ComplexSpecies([p[0], p[1], Species("S")], called_from_complex = True)
+
+    assert str(pc) == str(p) #these should have the same name!
+    assert str(p) == str(pc.polymers[0])
+    assert len(pc.complexes) == 0
+
+    pc2 = PolymerConformation(polymer = [Species("m1"), Species("m2"), Species("m3")])
+    assert pc2 == pc
```
dr3y, line 19: interesting. Can you make them reverse here as well?

Tests/test_combinatorial_complex.py
```diff
@@ -0,0 +1,268 @@
...
+    species_to_add = CC.compute_species_to_add(C, Complex([X, Y]))
+    assert species_to_add is None
+    #C contains more species than Complex([X, X])
+    species_to_add = CC.compute_species_to_add(C, Complex([X, X]))
+    assert species_to_add is None
+    #S is not in C
+    species_to_add = CC.compute_species_to_add(Species("S"), C)
+    assert species_to_add is None
+    species_to_add = CC.compute_species_to_add(Complex([Species("S"), X]), C)
+    assert species_to_add is None
```
dr3y, line 106: this is very cool. Not sure I completely understand it but it seems useful

biocrnpyler/components_basic.py
```diff
@@ -8,14 +8,41 @@
...
+    def length(self, length):
+        if length is None:
+            self._length = length
+        elif isinstance(length, int) and length >= 0:
+            self._length = length
+        else:
+            raise ValueError("Length must be a non-negative integer!")
 
-class DNA(Component):
+class DNA(LengthComponent):
```
dr3y, line 39: i guess `DNA_part` should be a `LengthComponent` too? I don't think it should be a `DNA` though

biocrnpyler/components_combinatorial_complex.py
```diff
@@ -0,0 +1,342 @@
...
+        #The order these run in is important!
+
+        #1. set final_states
+        self.final_states = final_states
+        #2. set initial_states
+        self.initial_states = initial_states
+        #3. set intermidiate_states
+        self.intermediate_states = intermediate_states
+        #4. set excluded_states
+        self.excluded_states = excluded_states
```
dr3y, line 86: could it be impossible to form the complex if certain states are excluded?

biocrnpyler/components_combinatorial_complex.py
```diff
@@ -0,0 +1,342 @@
...
+                    s_list.append(s)
+                    cs = Complex(s_list)
+
+                    #append the new combination if it isn't excluded
+                    if binder not in self.excluded_states and bindee not in self.excluded_states and cs not in self.excluded_states:
+                        combinations.append((binder, bindee, cs))
+                    #update bindee
+                    bindee = cs
+
+            return combinations
```
dr3y, line 239: I don't want to think about how this can be applied to integrases at the moment but I'm sure that when I get around to that I'll have more comments

biocrnpyler/components_combinatorial_conformation.py
```diff
@@ -0,0 +1,504 @@
...
+
+                active_complex = False
+                if any([s.get_complex(c) is not None for c in self.activating_complexes]):
+                    active_complex = True
+
+                innactive_complex = False
+                if any([s.get_complex(c) is not None for c in self.inactivating_complexes]):
+                    innactive_complex = True
+                    if (active_state and len(self.promoter_states)>0) or active_complex:
+                        warnings.warn("innactive_complex conflicts with active_complex or active_state in CombinatorialConformationPromoter. Defaulting to innactive.")
```
dr3y, line 497: typo in warning :)

biocrnpyler/components_combinatorial_conformation.py
```diff
@@ -0,0 +1,504 @@
...
+                        
+                if (active_state or active_complex) and not innactive_complex:
+                    self.dna_to_bind = s.polymers[0][self.promoter_location]
+                    self.name = self._get_part_id(s) #Reset name for unique addressable promoter states
+                    promoter_reactions += Promoter.update_reactions(self)
+        self.name = old_name
+        return promoter_reactions + conformation_reactions
```
dr3y, line 504: I don't think i've fully understood this but it seems useful

biocrnpyler/mechanisms_conformation.py
```diff
@@ -0,0 +1,37 @@
...
+    def update_reactions(self, s0, sf, additional_species = None, component = None, part_id = None, **kwords):
+
+        if part_id is None:
+            repr(s0)+"-"+repr(sf)
+
+        if additional_species is None:
+            additional_species = []
+
+        kf = component.get_parameter("kf", part_id = part_id, mechanism = self)
+        kr = component.get_parameter("kr", part_id = part_id, mechanism = self)
```
dr3y, line 35: so there is no default "k" to use for this? I am just wondering if i specify "kf" in parameters (nonspecifically) i guess that would cause a lot of problems

biocrnpyler/utils.py
```diff
@@ -127,6 +129,9 @@ def member_dictionary_search(member,dictionary):
                         warn(f"dictionary search output was {out_value} but now it will be {cur_value}")
                     out_value = cur_value
 
+            if 'reaction' in dictionary and out_value is None:
+                out_value = dictionary['reaction']
```
dr3y, line 133: i think i disagree with having 'reaction' be a special word for this search function. It is supposed to be a generic dictionary search function and not specifically for use with plotting

dr3y (changes requested):
I didn't read super deeply in combinatorial conformation/complex but i did offer a few comments, if you want to take a look. Main thing I oppose would be using "reaction" as a special word in `member_dictionary_search`

biocrnpyler/components_combinatorial_conformation.py
WilliamIX: The idea is that I reuse a Promoter by changing its internal "DNA_to_bind" (instead of enumerating many promoters like you do in DNA construct). This allows many species (conformations) to reuse the same promoter object. It is a little hacky, but because all the promoters make the same ultimate transcript I think it is okay.

biocrnpyler/mechanisms_conformation.py
WilliamIX: kf = kr = 100 for instance would allow for fast transfer between different complexes. I also built in a way to rename the part_id for each conformation (self._get_part_id(s)) to make parameter files cleaner because the strings used to represent polymer conformations are gross.

biocrnpyler/components_combinatorial_complex.py
WilliamIX: I didn't think about that either.....but it might solve some of the binding enumeration problems if you integrate it into an IntegraseEnumerator in a clever way!

biocrnpyler/components_basic.py
WilliamIX: Sure, I was unaware of LengthComponents but that makes sense

Tests/Unit/test_polymer_conformation.py
WilliamIX: What do you mean by that? that a polymer and its reverse are considered equal?

The reason for this is that, in order to keep code sane, I basically decided that either one would add complexes to a PolymerConformation or to a Polymer. This whole thing with binding a complex in a Conformation and updating the internal polymer lead to a ton of issues regarding the order of binding operations resulting in different names (I believe we had a good conversation about this a few months ago). Instead, I restricted the scope: if you want to use Conformations, the internal Polymer is immutable. However, that means a Conformation without anything should be the same as the underlying Polymer....

Tests/Unit/test_complex.py
WilliamIX: Complex will automatically make a PolymerConformation from another PolymerConformation. However, it does not make PolymerConformations from Polymers automatically. This is because once in a PolymerConformation, the internal Polymer is immutable. This means that the order Complex is applied matters when mixing PolymerSpecies with PolymerConformations. To avoid this, users who want to use Conformaitons should just place an immutable Polymer inside and use Conformations for everything.

.github/codecov.yml
WilliamIX: After today, it should not be changed...but I had to add it this branch to get it to pass coverage because of the changes to testing.

biocrnpyler/utils.py
WilliamIX: ok, we can remove that check here and just do it manually in plotting

WilliamIX: resolved on next commit.

setup.py
WilliamIX: good point - fixed on next commit.

dr3y (dismissed):

biocrnpyler/__init__.py
```diff
@@ -15,6 +15,8 @@
 from .dna_part_terminator import *
 from .construct_explorer import *
 from .integrase_enumerator import *
+from .components_combinatorial_complex import *
+from .components_combinatorial_conformation import *
```
ayush9pandey, line 20: I have a general concern here, not directly related to this PR. Should we be importing all classes and functions in `__init__.py`? Until now we have been importing each and everything here but now I am wondering if this is a good software engineering practice. Anyone has any thoughts on this @zoltuz @WilliamIX @dr3y ?

biocrnpyler/components_basic.py
```diff
@@ -8,14 +8,41 @@
 from .reaction import Reaction
 from .species import Complex, Species
 
+class LengthComponent(Component):
+    """
+    A high level class for Components with Length
```
ayush9pandey, line 13: So now all components will actually be "LengthComponent" objects? I'm not sure how I feel about that. Like in the test below, an RNA component is a "LengthComponent" object. This makes things confusing for the user who only works with the basic biocrnpyler classes. Is there a way around it so that every component is still a `Component` object?

ayush9pandey: Why couldn't length just be an optional attribute of the `Component` class. Do we need this super class? I don't like the fact that DNA, RNA, etc are now "LengthComponent".

ayush9pandey (changes requested):
I looked at the tests and the jupyter notebook briefly. Things seem like they are working (I don't understand it all) but I have some concerns as I commented particularly about the super class for Components.

biocrnpyler/__init__.py
WilliamIX: I've thought about this too....I think it might be good (but a pain in the ass) to refactor the biocrnpyler file structure to something like:

- BioCRNpyler
- - Mixture
- - - mixture.py, mixtures_cell.py etc.
- - Component
- - - component.py, components_basic.py etc
- - Mechanisms
etc

I think having "From biocrnpyler import *" grab all of these is then great, but that also gives easy leeway to do things like from biocrnpyler.mixture import Mixture if we want to be more careful about it.

Lets not make that part of this PR though.

biocrnpyler/components_basic.py
WilliamIX: I can't remember why I (or maybe someone else) did this - happy to remove LengthComponent.

WilliamIX: @ayush9pandey removed LengthComponent

ayush9pandey (approved):

PR #254, opened 2021-07-18
Dev to Master for Debugging

WilliamIX:


WilliamIX:
@ayush9pandey @zoltuz following up from my slack message the build seems to fail even on Dev (which successfully built before). This means some version of some software (probably jupyter?) got upgraded and no longer works the way we expected. Any thoughts on how to fix this?

PR #257, opened 2021-08-05
Update component.py to allow compartment attribute

zjuradoq:
Added compartment attribute to Component under set_species.


zjuradoq:
Was submitted to wrong branch.

PR #258, opened 2021-08-05
Added compartment to component.py

zjuradoq:
Added compartment attribute to Component under set_species.


ayush9pandey:
@zjuradoq I think this PR was for the compartment issue in component.py, so it would be better if we could move the membrane component related changes to a different PR as that is a new feature whereas the other thing is a bug fix.

zjuradoq:
Will re-request pull request in parts to better manage changes to dev branch.

PR #264, opened 2022-01-08
Renamed files and created an index "0 Tutorial Index.txt" via the scipt "generate_index.py"

WilliamIX:
This PR created an index of the different BioCRNpyler classes and where they occur in the example notebooks. To do this, a few example notebooks were renamed and moved around.


WilliamIX:
Note: all the test seem to run, but coverage is failing for some reason. Am looking into it.

ayush9pandey:
The PR file changes look good to me. We can approve it - I don't think codecov will fail because of these changes so even if the issue with codecov is not fixed we might go for the merge.

ayush9pandey (dismissed):

ayush9pandey (approved):

PR #267, opened 2022-03-03
Using py 3.7, 3.8, and 3.9 instead of 3.6, 3.7, and 3.8

WilliamIX:
This change is necessary because 3.6 is no longer supported PyO3 (pywinpty) which is one of our dependencies.

### DIFF
--- a/.github/workflows/install_and_code_cov.yml
+++ b/.github/workflows/install_and_code_cov.yml
@@ -7,7 +7,7 @@ jobs:
         max-parallel: 4
         matrix:
           os: [windows-latest, ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


WilliamIX:
@ayush9pandey just keeping you in the loop. This morning, all of the testing starting failing. Turns out the issue was python 3.6 is no longer supported by one of our dependencies' dependencies. I tried adding 3.9 as a version, but that is not compatible with ForceAtlas2 (at a minimum - and potentially other packages as well). Interim fix is to only test on 3.7 and 3.8.

ayush9pandey:
This is obsolete now with #276 and #277

PR #276, opened 2024-06-19
updates for Python 3.12 compability

murrayrm:
This PR includes a number of small updates to allow bioCRNpyler to install properly on Python 3.12 (also tested on Python 3.8).  The changes are all small and reflect updates in `fa2`, `bokeh`, and `nbconvert`.

Note that numpy 2.0 also causes problems.  Since this just came out 3 days ago, for now I have pinned numpy < 2.0 in `setup.py`

A bit more info:
* fa2: the original fa2 is no longer maintained and fails after Python 3.8.  A modified version is available in [fa2_modified](https://github.com/AminAlam/fa2_modified)
* Fixed an error in a `randint` call were `1e6` was being interpreted as a float (so I explicitly converted it to an `int`).
* The current version of bokeh uses `width` to set the width instead of `plot_width`.
* There were some small issues in unit test infrastructure in the way that warnings were handled and some changes in how exceptions are declared that I fixed.

Unit tests on my local machine all pass if `bioscrape` is installed (if not, there is a bug in `examples/Specialized\ Tutorials/1.\ CombinatorialPromoter\ Details.ipynb` that causes a failure).

It would be useful if this PR could be checked, merged, and pushed to PyPI in the next few days, so that bioCRNpyler is working when I talk about it at SEED (on 25 Jun 2024...).

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ Install with all optional dependencies::
 
     $ pip install biocrnpyler[all]
 
+(Note that on some operating systems you made need to use "\[all\]" to avoid shell errors.)
+
 Further details about the installation process can be found in the [BioCRNPyler wiki](https://github.com/BuildACell/BioCRNPyler/wiki#installation).
 
 # Bugs
--- a/Tests/Combinatorial/test_jupyter_notebooks.py
+++ b/Tests/Combinatorial/test_jupyter_notebooks.py
@@ -2,7 +2,7 @@ import pytest
 nb_not_installed = False
 try:
     import nbformat
-    from nbconvert.preprocessors import ExecutePreprocessor
+    from nbconvert.preprocessors import ExecutePreprocessor, CellExecutionError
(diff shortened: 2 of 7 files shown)


ayush9pandey:
Thanks for fixing this! I just ran into the fa2 incompatibility issue with biocrnpyler last week too. Will take a look now.

ayush9pandey:
Biocrnpyler installs fine on 3.12 with this PR and all tests pass without errors. 

I will create another PR to update the package installation setup with pyproject.toml so that PyPi distribution can be updated.

ayush9pandey (approved):

WilliamIX:
@murrayrm and @ayush9pandey thanks for making these changes! I'd been considering if we needed up update BioCRNpyler at some point for newer python versions as well. When I last looked into it, FA2 was the major issue so I'm glad you were able to find a new version that works.

PR #281, opened 2024-08-19
Update render_network_bokeh to use Scatter instead of Square + other small fixes

murrayrm:
The main change in this PR is to fixe a problem with render_network_bokeh, which was trying to use the old `Square` function in bokeh.models.  This has been replaced with `Scatter`, as documented [here](https://docs.bokeh.org/en/2.4.1/docs/reference/models/markers.html).

Also added a unit test and fixed a couple of docstrings that need to be raw to avoid errors with backslashes (there are probably more of these that need fixing...).

In addition, I found some scripts that did not run in the latest version of the code, so I modified those to allow them to run.

Note: `1. CombinatorialPromoter Details.ipynb` requires the `seaborn` package to run.  This is not currently included in the dependencies.  I didn't fix that here, but it would be good to do that at some point.


ayush9pandey (approved):
Everything looks good. We'll need to make the docstring change at other places in the future.

ayush9pandey:
Ready to merge.

ayush9pandey:
Merging.

PR #294, opened 2025-04-05
create Read the Docs documentation

murrayrm:
This PR creates a User Guide that can be viewed on Read the Docs.  Once this PR is merged we will need to set things up on readthedocs.org, but you can view what the output of the PR is [here](https://murrayrm-biocrnpyler.readthedocs.io/en/readthedocs-26mar2025/index.html).  The intent of this PR is to serve as a starting point for building out a more complete (and readable) User Guide.

Changes:
* Rearranged the `docs/` directory to use a top-level directory for documentation sources rather than separate `source` and `build`.  (I didn't see any need for the extra layer of folders before getting to the documentation files.)
* Created a basic introduction to bioCRNpyler based on an early two page summary of the package.  We can expand this later, but I wanted to have something simple in place.
* Created a "tutorial" chapter that right now consists of sections containing each of the tutorials from the bioCRNpyler `examples` directory.  I think we should do a separate single notebook tutorial at some point, but putting in these notebooks as sections within this chapter seemed like an OK starting point.
* Created a "reference manual" chapter that right now users `autosummary` to generate pages for all bioCRNpyler functions.  This is a bit ugly since it documents everything by modules (which are not really visible to the user), but again I wanted something simple as a start.
* Updated the section delimiters in the example Jupyter notebooks so that each file has a single top-level heading (#) and any headings within the notebook are second level (##).
 
@zjuradoq Once this is merged, I suggest we create a chapter for the membrane/transport modeling work you are doing.  As a start, we could do something like the "tutorial" chapter and create a file `membrane_transport.rst` that has in it the various Jupyter notebooks you have created.  We can later expand that to provide a bit more of a narrative introduction followed by examples.


ayush9pandey (approved):
All changes look good @murrayrm 

For the membrane tutorials, is it possible to use the jupyter notebooks (similar to how other tutorials are linked right now)? Perhaps we can create another section for advanced tutorials / specialized tutorials to host the membrane toolbox tutorial.

For hosting the wiki on biocrnpyler.readthedocs.io, are there steps that I can help with?

ayush9pandey:
Merging...

Issue #298, opened 2025-05-21
Drop python 3.8 support for biocrnpyler?


ayush9pandey:
Python 3.8 is now EOL (https://peps.python.org/pep-0569/)
and most standard python packages have stopped supporting v3.8.

The one that affects biocrnpyler is `python-libsbml`. It seems that they dropped 3.8 for macos wheels for their latest release. See here:

https://pypi.org/project/python-libsbml/5.20.5/#files

This is why our test for v3.8 is failing for macos but passing for windows and ubuntu (for now!)

So, I suggest that we upgrade our minimum python version to 3.9 and also drop the testing for 3.8.

WilliamIX:
This should be OK.

ayush9pandey:
We have tagged releases: https://github.com/BuildACell/bioCRNpyler/releases so we can always go back to previous versions.

Issue #292, opened 2025-02-27
Generate list of available mechanisms, components, and mixtures in user-friendly format


ayush9pandey:
With an automated documentation setup, we should be able to include a webpage for biocrnpyler that shows all supported types of mechanisms, components, and more.

Issue #291, opened 2025-02-24
Option to show compartments with CRN pretty print


ayush9pandey:
With new developments in #286 (@zjuradoq), it would be good to allow CRN pretty print to also show compartments of species.

Issue #288, opened 2025-02-12
Compartments in BioCRNpyler: updates required


ayush9pandey:
Related to #287: 
- Each species should have a compartment.
- Each component should have the same compartment as the species that it contains
- Each mechanism should either be denoted as a inter-compartmental mechanism or a same compartment mechanism. For the latter, all new complexes that the mechanism creates should use the same compartment as the species that it uses. For the former, the compartment should be named explicitly (a membrane compartment for example)
- Each mixture should have the same compartment as its species.

PR #299, opened 2025-05-24
Refactor biocrnpyler modules and bug fixes

ayush9pandey:
## Why this PR?
This PR proposes a refactor for modules within biocrnpyler. As the package has grown, all source code has collected over the years under `biocrnpyler/` directory with no structure to any of the files within the source code. This impacts the way users import classes, and at the moment it's all part of "biocrnpyler" with no distinction between whether something is a mechanism or a component, or a mixture. Additionally, for future developers who would like to contribute a single mechanism, it is difficult to know exactly where to add such a contribution. So, with this PR, I would like to propose a revised file structure:

```
biocrnpyler/
├── __init__.py
├── core/
│   ├── __init__.py
│   ├── mixture.py
│   ├── component.py
│   ├── species.py
│   ├── chemical_reaction_network.py
│   ├── compartment.py
│   ├── parameter_database.py
│   └── ... (other core modules: the main classes of biocrnpyler)
├── components/
│   ├── __init__.py
│   ├── basic.py
│   ├── membrane.py
│   ├── dna
│   ├──├──assembly.py
│   ├──├──CDS.py
│   ├──├──(all other DNA related components)
│   ├── ... (all other components will go in this library)
├── mixtures/
│   ├── __init__.py
│   ├── cell.py
│   ├── extract.py
├── mechanisms/
│   ├── __init__.py
│   ├── binding.py
│   ├── signaling.py
│   ├── transport.py
│   ├── ... ((all other mechanisms will go in this library))
├── tests/ (same as before)
├── examples/ (same as before)
```

## Important note about importing modules

These changes are backward compatible. So, all other ways of importing modules that worked previously will continue to work. For example, the wildcard import `from biocrnpyler import *` still works, and users can still do `from biocrnpyler import Species, TxTlExtract, DNA`. Although both of these previous import statements are confusing in terms of representing the functionality. The updated structure will allow a user to instead import:
```
from biocrnpyler.core import Species
from biocrnpyler.mixtures import TxTlExtract
from biocrnpyler.components import DNA
```
thus, clarifying that DNA is a component and TxTlExtract is a mixture. If three lines of imports is not desirable, the one line import that was used previously will continue to work too.

## New feature: Compartment update during compile_crn
One of the main features in this PR enables the user to run the following during the compilation of CRN (usually, the last step in model building):

```
# define all components, mechanisms, and Mixture M
component1 = Protein("gfp") # by default this is placed in the default compartment
membrane_component = IntegralMembraneProtein(...) 
# the membrane components and mechanisms usually automatically place things in internal and external compartments 
membran_mech = Simple_Transport(...)
# Since component1 was defined without being specific about the compartment, the following
# can be used to update all "default" compartment to a new_compartment
# get the desired compartment
new_compartment = membrane_component.product.compartment
CRN = M.compile_crn(compartment = new_compartment)
```

This feature will also be useful to update compartments of a component after a CRN is compiled already, and we decide to compile again with the updated compartments. 

This description is also added to the docs/intro.rst so that users don't have to refer to this PR in the future :) 

## New feature: Consistency with Compartment getter/setters in Components

In earlier versions, compartment attribute to compartments, species, and mixtures were not consistent. This PR adds the getter and setter functions needed to be consistent with compartments. 

## Bug Fixes

This structure along with the recent readthedocs support from #294 can now address #292 easily (to provide list of all mechanisms and components).

In the process of updating this structure, I have fixed bugs: #298, #288, #287 , #291 , #307

Since this is a large-ish PR, it would be great if active biocrnpyler users/developers could give their opinions @zjuradoq @murrayrm @WilliamIX @hariKRN2000


WilliamIX:
@ayush9pandey thanks for taking initiative on this! 

I generally approve of this refactor, but want to make sure we don't take anyone who might be using the software as a dependency by surprise when all their imports fail.

Is there any way we can detect the old input style and give a useful error so people can understand what happened?

ayush9pandey:
> However, in the following code block, after adding a DNA assembly, RNAP is not found as shown by the reported initial condition of `protein[RNAP] (@ 0).` I believe this is also occurring for other species such as `metabolite[Fuel_3PGA] (@ 0)`. I ran a test and changed the `param_name` in `txtl_toolbox_parameters.txt` from `RNAP` to `protein_RNAP`, which "fixed" the initial condition error. Lastly, this error in the parameter search may have existed in the `master` branch since `RNAse` in `txtl_toolbox_parameters.txt` is set to 20.2 but is not "found" and thus set to 0 for the CRN example.
> 
> I believe there might be an issue when building the `mixture` and the `param_name` that is being searched.

Thank you for finding this @zjuradoq !!

murrayrm:
One other thought: what are we doing for version numbering in bioCRNplyer?  Presumably we want to use [semantic versioning](https://semver.org) and bump up at least the minor version number if not the major version number.

ayush9pandey:
> One other thought: what are we doing for version numbering in bioCRNplyer? Presumably we want to use [semantic versioning](https://semver.org) and bump up at least the minor version number if not the major version number.

The semantic version is specified in our [pyproject.toml](https://github.com/BuildACell/bioCRNpyler/blob/master/pyproject.toml) file. I bumped the minor version, so from 1.2 to 1.3, in this PR ([here](https://github.com/BuildACell/bioCRNpyler/blob/refactor-modules/pyproject.toml)). With all the changes we are making this month (membrane models, new docs, this one), it might make sense to bump the major version. I'm open to other thoughts.

Once the PR is merged, I can run the [pypi-release](https://github.com/BuildACell/bioCRNpyler/blob/master/.github/workflows/pypi-release.yml) action to update the pypi distribution.

biocrnpyler/__init__.py
```diff
@@ -1,47 +1,46 @@
...
+from .components.dna.terminator import *
+from .components.construct_explorer import *
+from .components.integrase_enumerator import *
+from .components.combinatorial_complex import *
+from .components.combinatorial_conformation import *
+from .components.component_enumerator import *
+from .components.membrane import *
 
-from .chemical_reaction_network import *
-from .component import *
```
murrayrm, line 5: Rather that importing from each subdirectory from here, would it be possible to use a more recursive import structure?  So in this file you do `from .components import *` and then in `components/init.py` you import from `basic`, etc?  That would allow things to be more modular, since developers would not have to make changes to the global package when they are added a new component, mechanism, etc.

biocrnpyler/core/chemical_reaction_network.py
```diff
@@ -47,7 +47,7 @@ def species(self):
     def species(self, species):
         if not hasattr(self, "_species"):
             self._species = []
-            self._species_dict = {}
+            self._species_set = set()
```
murrayrm, line 50: It would be useful to document the change of the species dict object to a set object in the PR description, so that we can get more information about why this change was made.

biocrnpyler/core/chemical_reaction_network.py
```diff
@@ -65,7 +65,7 @@ def reactions(self, reactions):
             raise AttributeError("The reactions in a CRN cannot be removed or modified. New reactions can be added with CRN.add_reactions(...).")
 
     
-    def add_species(self, species, copy_species = True):
+    def add_species(self, species, copy_species = True, compartment = None):
```
murrayrm, line 68: PEP8: as long as you are making a change here (and elsewhere?), you might as well use PEP8 rules (no spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter).

biocrnpyler/core/chemical_reaction_network.py
```diff
@@ -85,11 +85,15 @@ def add_species(self, species, copy_species = True):
...
+            if s not in self._species_set: #Do not add duplicate Species
+                if compartment is not None and s.compartment.name == "default":
+                    s.compartment = compartment
+                self._species_set.add(s)
                 self._species.append(s) #copy the species and add it to the CRN
+                
 
-    def add_reactions(self, reactions: Union[Reaction,List[Reaction]], copy_reactions = True, add_species = True) -> None:
+    def add_reactions(self, reactions: Union[Reaction,List[Reaction]],
+                      copy_reactions = True, add_species = True, compartment = None) -> None:
```
murrayrm, line 96: PEP8: spacing

biocrnpyler/core/chemical_reaction_network.py
```diff
@@ -115,7 +119,9 @@ def add_reactions(self, reactions: Union[Reaction,List[Reaction]], copy_reaction
 
                 # add all the Species in the reaction to the CRN
                 reaction_species = list(set([w.species for w in r.inputs + r.outputs]))
-                self.add_species(reaction_species, copy_species = copy_reactions)
+                self.add_species(reaction_species,
+                                 copy_species = copy_reactions,
+                                 compartment = compartment)
```
murrayrm, line 124: PEP8: spacing

biocrnpyler/core/component.py
```diff
@@ -63,6 +65,11 @@ def __init__(self, name: Union[str, Species],
         else:
             self.set_mixture(None)
 
+        if compartment is not None:
+            if not isinstance(compartment, Compartment):
+                raise TypeError("The compartment must be a Compartment object")
+            self.compartment = compartment
+
```
murrayrm, line 72: In other places (like `Species`), we allow the compartment to be a string.  Shouldn't that be allowed here as well (and then it gets converted internally to a `Compartment` object?).

biocrnpyler/core/component.py
```diff
@@ -85,21 +93,54 @@ def initial_concentration(self, initial_concentration):
     
         self._initial_concentration = initial_concentration
 
+    @property
+    def compartment(self):
```
murrayrm, line 97: I notice a lot of changes related to compartments. Can you give a short description of the changes in the PR description? Looks like you are adding functionality?

biocrnpyler/core/mixture.py
```diff
@@ -3,23 +3,24 @@
...
-from .utils import remove_bindloc
+from .compartment import Compartment
+from ..utils.general import remove_bindloc
 
 
 class Mixture(object):
-    def __init__(self, name="", mechanisms=None, components=None, parameters=None, parameter_file=None,
-                 global_mechanisms=None, species=None, initial_condition_dictionary=None, \
-                 global_component_enumerators=None,global_recursion_depth=4, local_recursion_depth = None, **kwargs):
+    def __init__(self, name="", mechanisms=None, components=None, parameters=None, compartment: Compartment = None,
```
murrayrm, line 21: `Compartment` or `str`?

biocrnpyler/core/mixture.py
```diff
@@ -362,33 +385,58 @@ def get_initial_concentration(self, S: Union[List, Species], component=None):
...
+        # 5) return the updated species list as before
+        return all_species
+        # self.crn.add_species(new_species, copy_species = copy_species,
+        #                      compartment = compartment)
+
+        # if not no_initial_concentrations:
+        #     init_conc_dict = self.get_initial_concentration(remove_bindloc(new_species), component)
+        #     self.crn.initial_concentration_dict = init_conc_dict
+
+    def apply_global_mechanisms(self, species, compartment = None) -> Tuple[List[Species], List[Reaction]]:
```
murrayrm, line 426: Describe in PR description?

docs/_autogen_components.rst
murrayrm, line 1: I'm working on a PR that builds on this one and will automatically generate the `_autogen_*` files from the source when making the documentation.  When I post that change, I'll remove this file.

murrayrm (commented):
Overall these changes look very good to me (nice work!).  There are a couple of places where it looks like some new functionality and/or refactoring is being done, and it would be nice to document these in the PR description.

There is also lots of PEP8 non-compliant code (mainly spacing around parameter assignments). Probably not worth fixing here, but I would encourage using PEP8 for future changes.

I also found a few other things I wasn't sure about, so worth taking a look at those and fixing (+ adding unit tests?) as needed.

I think we should try to merge this fairly soon, so that others (including me) can build on it.

ayush9pandey:
Thanks for all the suggestions @murrayrm !

I will fix these and will add more descriptions to document the changes. There's a bug with setting parameter values in some mixtures that I haven't quite tracked down yet, so progress on this PR was slowed. I will get to it in a day or two though..

murrayrm:
Separating out the two changes is "cleaner", but if it is going to be a big hassle then it seems fine to just fix them both up in this PR.  Either way, we'll want both sets of changes in place (I'm going to be using compartments quite a bit for synthetic cell modeling this summer...).

biocrnpyler/core/chemical_reaction_network.py
ayush9pandey: Thanks for pointing this out. I added the following to the docstring:
        A _species_set is used to ensure that no duplicate species are added to the CRN.
        In earlier BioCRNPyler versions, a _species_dict with key as species, and value as "True" was used, which was unintuitive.
        Since sets are designed to keep unique elements, so we use a set to keep track of species.

biocrnpyler/core/chemical_reaction_network.py
ayush9pandey: Yes, will use autopep8 formatter to all files.

biocrnpyler/core/chemical_reaction_network.py
ayush9pandey: Should be resolved now.

biocrnpyler/core/chemical_reaction_network.py
ayush9pandey: Should be resolved now.

biocrnpyler/core/component.py
ayush9pandey: You are right -- this is one of the main features that I have added in this PR. Will add more details to the PR description.

biocrnpyler/core/mixture.py
ayush9pandey: Will do.

biocrnpyler/core/component.py
ayush9pandey: Usually, a component out of the core/components/ library will create the Component class and should handle the compartment creation for the component (like IntegralMembraneProtein handles its compartments and allows strings).

biocrnpyler/core/mixture.py
ayush9pandey: Done.

docs/_autogen_components.rst
ayush9pandey: Sounds good.

ayush9pandey:
@murrayrm I have addressed most comments now and have pushed all changes. Thank you so much for your detailed review! Would you please quickly review again and confirm if it is OK to merge?

murrayrm (dismissed):
Read through the latest set of changes and I think this is good to merge.

ayush9pandey:
Merged. Thank you for your reviews, suggestions, and help with fixing various things on this PR! @murrayrm @zjuradoq @WilliamIX

Issue #306, opened 2025-08-09
Update compartment handling in membrane models?


murrayrm:
It looks like the way that membrane models handles compartments (in branch `refactor-modules`) is not quite right.

Consider the following code:
```
import biocrnpyler as bcp

# Create compartments for the inside and outside of the detector cell
detector_internal = bcp.Compartment(name='DetectorInternal')
detector_membrane = bcp.Compartment(name='DetectorMembrane')
external = bcp.Compartment(name='External')

# Define membrane protein
alphaHL = bcp.IntegralMembraneProtein(
    'alphaHL_monomer', product='alphaHL', size=7, 
    compartment=detector_internal, membrane_compartment=detector_membrane)
alphaHL_channel = bcp.MembraneChannel(
    alphaHL.product, substrate='atc', 
    internal_compartment=detector_internal, external_compartment=external)

# Integration and transport mechanisms
mech_integration = bcp.Membrane_Protein_Integration()
integration_mechanisms = {mech_integration.mechanism_type: mech_integration}
mech_transport = bcp.Simple_Transport()
transport_mechanisms = {mech_transport.mechanism_type: mech_transport}

# Create a mixture
mixture = bcp.SimpleTxTlExtract(
    'DetectorCell',
    components=[alphaHL, alphaHL_channel],
    mechanisms=integration_mechanisms | transport_mechanisms,
    parameter_file='membrane_toolbox_parameters.txt',
)

crn = mixture.compile_crn()
print(crn.pretty_print(show_compartments=True))
```

The (partial, reformatted) output is
```
Species(N = 5) = {
    complex[7x_protein[alphaHL_monomer] in default.] (@ 0),
    atc in External. (@ 0),
    atc in DetectorInternal. (@ 0),
    protein[alphaHL_monomer in DetectorInternal.] (@ 0),
    protein[alphaHL in default.(Passive)] (@ 0),
}
```

This shows `complex[7x_protein[alphaHL_monomer]]` and  `protein[alphaHL]` in compartment 'default', whereas I think `7x_protein[alphaHL_monomer]` should be in 'DectectorMembrane' and `protein[alphaHL_monomer]` should be in 'DetectorInternal'.

Also, why does `IntegralMembraneProtein` using keywords `compartment` and `membrane_compartment` while `MembraneChannel` uses keywords `internal_compartment` and `external_compartment`?

@zjuradoq: Can you comment on this when you get a chance?  Am I using the keywords properly and interpreting the output properly?

[...]

PR #312, opened 2025-08-13
Patch compartment assignments in membrane component

zjuradoq:
A patch to fix compartment assignments specifically in the membrane component `IntegralMembraneProtein.`


ayush9pandey (dismissed):
I tested the code in #306 and it looks like it works as expected with this fix.

murrayrm:
This PR doesn't quite fix the issue. Consider the following code:
```
import biocrnpyler as bcp

# Create compartments for the inside and outside of the detector cell
detector_internal = bcp.Compartment(name='DetectorInternal')
detector_membrane = bcp.Compartment(name='DetectorMembrane')
external = bcp.Compartment(name='External')

# Define membrane protein
alphaHL = bcp.IntegralMembraneProtein(
    'alphaHL_monomer', product='alphaHL', size=7, 
    compartment=detector_internal,  # THIS SHOULD BE `internal_compartment`
    membrane_compartment=detector_membrane)
alphaHL_channel = bcp.MembraneChannel(
    alphaHL.product, substrate='atc', 
    internal_compartment=detector_internal,  # IS THIS NEEDED?
    external_compartment=external)

# Integration and transport mechanisms
mech_integration = bcp.Membrane_Protein_Integration()
integration_mechanisms = {mech_integration.mechanism_type: mech_integration}
mech_transport = bcp.Simple_Transport()
transport_mechanisms = {mech_transport.mechanism_type: mech_transport}

# Create a mixture
mixture = bcp.SimpleTxTlExtract(
    'DetectorCell',
    components=[alphaHL, alphaHL_channel],
    mechanisms=integration_mechanisms | transport_mechanisms,
    parameter_file='membrane_toolbox_parameters.txt',
)

crn = mixture.compile_crn()
print(crn.pretty_print(show_compartment=True))
```
Comments in CAPS above identify items that I think need an update (more below).

The output from this command is:
```
Species(N = 5) = {
    complex[7x_protein[alphaHL_monomer] in default.] (@ 0),
    atc in External. (@ 0),
    atc in DetectorInternal. (@ 0),
    protein[alphaHL_monomer in default.] (@ 0),
   protein[alphaHL in DetectorMembrane.(Passive)] (@ 0),  
}
```
Note that `protein[alphaHL_monomer]` and `7x__protein[alphaHL_monomer]` are still in default (instead of `DetectorInternal`).

In terms of the comments in the code:
* The `IntegralMembraneProtein` component still uses `compartment` to specify the internal component.  I think it would be more consistent with the naming in this module to use `internal_component` (and `membrane_component`) in this constructor.
* The `MembraneChannel` component gets the membrane channel from the `membrane_protein` (first argument), but determines the internal compartment from the `compartment` keyword.  Two issues here:
** If we keep the `compartment` keyword to specify the internal compartment, I think we should rename it as `internal_compartment`.
** Why are we pulling the membrane compartment from `membrane_protein` but not the internal compartment?  It seems like we could (should?) pull this from `membrane_protein.membrane_protein.compartment`?  But perhaps there are cases where the membrane protein only has information about the membrane compartment?

Also, a few other comments on this module:
* There is some nonstandard spacing of comments in the code:
   ```
   # PROTEIN
       if isinstance(integral_membrane_protein, str):
           ...
   ```
   should be
   ```
   # PROTEIN
   if isinstance(integral_membrane_protein, str):
       ...
   ```

* The naming scheme for mechanism classes uses underscores instead of camel case (`Simple_Diffusion` instead of `SimpleDiffusion`).

zjuradoq:
@murrayrm if you run the lines of code @ayush9pandey had shared, the output is:
```
Species(N = 5) = {
complex[7x_protein[alphaHL_monomer] in DetectorInternal.] (@ 0),  atc in External. (@ 0),  atc in DetectorInternal. (@ 0),  protein[alphaHL_monomer in DetectorInternal.] (@ 0),  protein[alphaHL in DetectorMembrane.(Passive)] (@ 0),  
}

Reactions (3) = [
0. 7protein[alphaHL_monomer] <--> complex[7x_protein[alphaHL_monomer]]
 Kf=k_forward * protein_alphaHL_monomer_DetectorInternal^7
 Kr=k_reverse * complex_protein_alphaHL_monomer_7x_DetectorInternal_
  k_forward=0.002
  found_key=(mech=membrane_protein_integration, partid=None, name=kb_oligomer).
  search_key=(mech=membrane_protein_integration, partid=alphaHL_monomer, name=kb_oligomer).
  k_reverse=2e-10
  found_key=(mech=membrane_protein_integration, partid=None, name=ku_oligomer).
  search_key=(mech=membrane_protein_integration, partid=alphaHL_monomer, name=ku_oligomer).

1. complex[7x_protein[alphaHL_monomer]] --> protein[alphaHL(Passive)]
 Kf = k complex[7x_protein[alphaHL_monomer]] / ( 1 + (protein[alphaHL(Passive)]/K)^4 )
  k=10.0
  found_key=(mech=membrane_protein_integration, partid=None, name=kex).
  search_key=(mech=membrane_protein_integration, partid=alphaHL_monomer, name=kex).
  K=0.5
  found_key=(mech=membrane_protein_integration, partid=None, name=kcat).
  search_key=(mech=membrane_protein_integration, partid=alphaHL_monomer, name=kcat).
  n=4

2. atc+protein[alphaHL(Passive)] <--> atc+protein[alphaHL(Passive)]
 Kf=k_forward * atc_DetectorInternal * protein_alphaHL_Passive_DetectorMembrane
 Kr=k_reverse * atc_External * protein_alphaHL_Passive_DetectorMembrane
  k_forward=0.1
  found_key=(mech=simple_membrane_protein_transport, partid=None, name=k_trnsp).
  search_key=(mech=simple_membrane_protein_transport, partid=alphaHL, name=k_trnsp).
  k_reverse=0.1
  found_key=(mech=simple_membrane_protein_transport, partid=None, name=k_trnsp).
  search_key=(mech=simple_membrane_protein_transport, partid=alphaHL, name=k_trnsp).
```

The main difference between the previous code shared by Ayush and the one you most recently ran is the last line, which indicates the compartment of the SimpleTxTlExtract. 
```
# Create a mixture
mixture = bcp.SimpleTxTlExtract(
    'DetectorCell',
    components=[alphaHL, alphaHL_channel],
    mechanisms=integration_mechanisms | transport_mechanisms,
    parameter_file='/content/BioCRNPyler/examples/Specialized Tutorials/membrane_toolbox_parameters.txt',
    compartment=detector_internal # <---- the line to change
)
```
Addressing your point, indicated by the CAPS:

```alphaHL = bcp.IntegralMembraneProtein(
    'alphaHL_monomer', product='alphaHL', size=7, 
    compartment=detector_internal,  # THIS SHOULD BE "internal_compartment"
    membrane_compartment=detector_membrane)
```
I previously explained why I chose to keep the parameter as `compartment` to keep it more arbitrary. However, if protein expression is restricted to the inside of a vesicle, then I do not see an issue with changing it to `internal_compartment`.

```
alphaHL_channel = bcp.MembraneChannel(
    alphaHL.product, substrate='atc', 
    internal_compartment=detector_internal,  # IS THIS NEEDED?
    external_compartment=external)
```
The line in question is needed to indicate the compartment of the substrate if choosing to use specific names. The only species `MembraneChannel` input is the `integral_membrane_protein` (`alphaHL.product`), which has the compartment of `'DetectorMembrane'`.

Thoughts on your other comments:
- _The IntegralMembraneProtein component still uses compartment to specify the internal component. I think it would be more consistent with the naming in this module to use internal_component (and membrane_component) in this constructor._
We can make this change if it makes it more consistent. I believe it might have been changed previously, but I reverted the change while writing the patch. @ayush9pandey Do you have thoughts on using `internal_compartment` vs `compartment`.

- _The MembraneChannel component gets the membrane channel from the membrane_protein (first argument), but determines the internal compartment from the compartment keyword. Two issues here:_
** _If we keep the compartment keyword to specify the internal compartment, I think we should rename it as internal_compartment._
** _Why are we pulling the membrane compartment from membrane_protein but not the internal compartment? It seems like we could (should?) pull this from membrane_protein.membrane_protein.compartment? But perhaps there are cases where the membrane protein only has information about the membrane compartment?_

I am not sure I follow. The MembraneChannel component gets the membrane channel from the first argument, and this is the only species that has a specified compartment. Also, can you clarify if you are referring to the MembraneChannel (uses argument of integral_membrane_protein) or the IntegralMembraneProtein (uses argument of membrane_protein).

I can edit spacing with the next edits on the patch.

zjuradoq:
I understand the problem now, and I agree with the statement below.

> For the specific example that I gave above, there is no DNA and so now new species are created by the SimpleTxTlExtract mixture => I think all of the components should have a defined compartment => nothing should be left in None (or 'default'). Is that correct?

I found that in the IntegralMembraneProtein component, the specified compartment is being overwritten, resulting in the default compartment. As I was working on fixing it, I wanted input on what assumptions should be made or how we might want to address user situations. For the specific example above, the membrane protein is defined in the IntegralMembraneProtein component, thus all the species should exist in the indicated compartment `detector_internal.` 

However, what if you had previously defined:
```
# Define membrane protein
alphaHL_monomer= Species(name='alphaHL_monomer', compartment=detector_internal0) #<---define monomer species first

alphaHL = IntegralMembraneProtein(
    alphaHL_monomer, product='alphaHL', size=7,
    compartment=detector_internal, membrane_compartment=detector_membrane)

alphaHL_channel = bcp.MembraneChannel(
    alphaHL.product, substrate='atc',
    internal_compartment=detector_internal, external_compartment=external)
```
Should IntegralMembraneProtein use the compartment of `alphaHL_monomer` (i.e. detector_internal0) or overwrite the compartment to be detector_internal? The options are:

- If compartment of `alphaHL_monomer` if not `default`: 
  - Use the compartment of `alphaHL_monomer` --> Compartments = **DetectorInternal0**, External, DetectorMembrane
  - Change the compartment of `alphaHL_monomer` based on `alphaHL` --> Compartments = **DetectorInternal**, External, DetectorMembrane
  
Sample code currently attempting onto PR #312 (running into some errors at the moment):
 ```
def __init__(self, membrane_protein: Union[Species, str, Component],
                 product: Union[Species,str, Component], direction:str=None, size:int=None, 
                 compartment='Internal', membrane_compartment='Membrane', cell:Union[str,int]=None,
                 attributes=None, **keywords):
       
        self.membrane_protein = self.set_species(membrane_protein)

        if self.membrane_protein.compartment.name not in ('default', compartment.name):
          warnings.warn("Inconsistent compartment with IntegralMembraneProtein and membrane_protein. Defaulting to use IntegralMembraneProtein compartment.", UserWarning) #<--can change this error message to be more specific
          self.membrane_protein.compartment.name = compartment.name # uses membrane protein compartment going forward
          # compartment = self.membrane_protein.compartment.name  # uses compartment parameter going forward
        else: 
          self.membrane_protein.compartment.name = compartment.name 
```

**Comment by murrayrm** [comment]
Yes, this addresses my issues and now works as expected!

**Comment by ayush9pandey** [comment]
On the warning below, I want to add that we should assign the compartment object (not the name). So, `self.membrane_protein.compartment = compartment`. Right?

> def __init__(self, membrane_protein: Union[Species, str, Component],
>                 product: Union[Species,str, Component], direction:str=None, size:int=None, 
>                 compartment='Internal', membrane_compartment='Membrane', cell:Union[str,int]=None,
>                 attributes=None, **keywords):
>       
>        self.membrane_protein = self.set_species(membrane_protein)
> 
>        if self.membrane_protein.compartment.name not in ('default', compartment.name):
>          warnings.warn("Inconsistent compartment with IntegralMembraneProtein and membrane_protein. Defaulting to use IntegralMembraneProtein compartment.", UserWarning) #<--can change this error message to be more specific
>          self.membrane_protein.compartment.name = compartment.name # uses membrane protein compartment going forward
>          # compartment = self.membrane_protein.compartment.name  # uses compartment parameter going forward
>        else: 
>          self.membrane_protein.compartment.name = compartment.name 
> ```

**Comment by zjuradoq** [comment]
> On the warning below, I want to add that we should assign the compartment object (not the name). So, self.membrane_protein.compartment = compartment. Right?

This is preferred, then it can be coded this way. I had it originally written your suggested way using `compartment.name` but it did not pass the checks. It kept failing when the compartment parameter was not given, with the default as `compartment='Internal'` (using `compartment.name` does not work since a str does not have a .name attribute). Currently, there is a line to check if the given compartment is a str and store only the name if it is not. Another option would be to make the compartment from the str, allowing us to use the code you suggested. Would making the compartment object be more streamlined?

**Comment by zjuradoq** [comment]
I have clarified the logic around the compartments, in addition to editing the spacing. PR passes all checks and is ready for review.

Lastly the scenario,
```
detector_internal = bcp.Compartment(name='DetectorInternal')
detector_membrane = bcp.Compartment(name='DetectorMembrane')
external = bcp.Compartment(name='External')

alphaHL = bcp.IntegralMembraneProtein(
    'alphaHL_monomer', product='alphaHL', size=7, 
    compartment='DetectorInternal',
    membrane_compartment='DetectorMembrane')
alphaHL_channel = bcp.MembraneChannel(
    alphaHL.product, substrate='atc', 
    compartment=detector_internal,
    external_compartment=external)
```
return, 
Species(N = 5) = {
complex[7x_protein[alphaHL_monomer] in DetectorInternal.] (@ 0),  atc in External. (@ 0),  atc in DetectorInternal. (@ 0),  protein[alphaHL_monomer in DetectorInternal.] (@ 0),  protein[alphaHL in DetectorMembrane.(Passive)] (@ 0),  
}
...

I am surprised that the current code manages the scenario as the user might have imagined. It might be a nice feature for the time being.

**Review by ayush9pandey — APPROVED** [review]

**Comment by zjuradoq** [comment]
There is a unit test for the cell parameter and examples in the [8. Multicellular_Transport.ipynb](https://github.com/BuildACell/bioCRNpyler/blob/main/examples/Specialized%20Tutorials/8.%20Multicellular_Transport.ipynb) uses it. I can look into it more tomorrow, but you do make a good point on whether it is needed. I used it previously when modeling multiple vesicles. You are correct that it just changes the name; this might be where the error is coming now, since we are using the Compartment objects.

I can remove the parameter, and we can leave it to the user to define the compartment correctly. This method might be more straightforward than chasing the naming error that is occurring now and accounting for all scenarios.

**Comment by zjuradoq** [comment]
Executing the `cell` parameter as it was first intended may now be a source of errors, given the implementation of the compartment component. The parameter and all its dependencies have been removed. Example [8. Multicellular_Transport.ipynb](https://github.com/BuildACell/bioCRNpyler/blob/main/examples/Specialized%20Tutorials/8.%20Multicellular_Transport.ipynb) has been updated.

**Comment by ayush9pandey** [comment]
@murrayrm If the update looks good to you, we can merge.

File: biocrnpyler/components/membrane.py [hunk 1]
```diff
@@ -3,45 +3,41 @@
...
         # Substrate
         self.substrate = self.set_species(substrate, compartment=internal_compartment)
         self.product = self.set_species(substrate, compartment=external_compartment)
         
         # Name the component
-        if cell is not None:
-            name = self.substrate.name + '_' + str(cell)
-        else:
-            name = self.substrate.name
+        name = self.substrate.compartment.name
```
murrayrm, line 40: Shouldn't this be `self.substrate.name`?  It seems like we want to name the component after the substrate species and not after the compartment that it is in (consistent with what was done before).

biocrnpyler/components/membrane.py
```diff
@@ -124,13 +120,9 @@ def __init__(self, membrane_protein: Union[Species, str, Component],
...
             self.membrane_protein.size = size
         
-        # Indicates cell
-        self.product.cell= cell
         # Name the component
-        if cell is not None:
-            name = self.membrane_protein.name + '_' + str(cell)
-        else:
-            name = self.membrane_protein.name
+        name = self.membrane_protein.compartment.name
```
murrayrm, line 124: Same comment as above; should this be self.membrane_protein.name?

biocrnpyler/components/membrane.py
```diff
@@ -233,11 +216,10 @@ def __init__(self, integral_membrane_protein: Union[Species, str, Component],
                                                    attributes=attributes)
             else:
                 print('Membrane channel direction not found.')
+        
         # Name the component
-        if cell is not None:
-            name = self.integral_membrane_protein.name + '_' + str(cell)
-        else:
-            name = self.integral_membrane_protein.name
+        name = self.integral_membrane_protein.compartment.name
```
murrayrm, line 221: As above.

biocrnpyler/components/membrane.py
```diff
@@ -365,11 +338,10 @@ def __init__(self, membrane_pump: Union[Species, str, Component],
                 self.membrane_pump.ATP= 1
             else: 
                 self.membrane_pump.ATP = ATP
+        
         # Name the component
-        if cell is not None:
-            name = self.membrane_pump.name + '_' + str(cell)
-        else:
-            name = self.membrane_pump.name
+        name = self.membrane_pump.compartment.name
```
murrayrm, line 343: As above.

biocrnpyler/components/membrane.py
```diff
@@ -469,10 +432,8 @@ def __init__(self, membrane_sensor_protein: Union[Species, str, Component],
         self.waste = self.set_species('ADP',  material_type='small_molecule', compartment=internal_compartment,
                                       attributes=attributes)
         # Name the component
-        if cell is not None:
-            name = self.membrane_sensor_protein.name + '_' + str(cell)
-        else:
-            name = self.membrane_sensor_protein.name
+        name = self.membrane_sensor_protein.compartment.name
```
murrayrm, line 435: As above.

murrayrm (commented):
Overall these changes look good, but I had a question about the naming of the components.  Once that is resolved (either explanation or changes), we can merge.

biocrnpyler/components/membrane.py
ayush9pandey: Yes, it looks like it should be `substrate.name`. @zjuradoq are we missing something here?

zjuradoq: @murrayrm please refer to above, I was not able to type faster than your comment about failing the unit test.

murrayrm:
@zjuradoq I sent you a PR for your branch that fixes the problem above.

murrayrm:
I can look into the error; it's possible that the unit test is not specific enough. What error are you specifically referring to? I have accepted the PR request on my branch.

murrayrm (approved):
Latest changes seem to be working => LGTM.

ayush9pandey:
Specialized tutorial notebooks are also run as part of the unit testing. See L26 [here](https://github.com/BuildACell/bioCRNpyler/blob/main/Tests/Combinatorial/test_jupyter_notebooks.py)
so it's strange that something failed on your local and was not flagged on the tests.

PR #311, opened 2025-08-13
Update sphinx documentation (User Guide + Reference Manual)

murrayrm:
This PR restructures the sphinx (readthedocs) documentation for the package, dividing the documentation in two parts:

* A "User Guide" that contains chapters on the main elements of the package and that provides a narrative description of how everything works.  There is also a chapter containing all of the tutorials.
* A "Reference Manual" that has documentation of all functions and classes in the package, as well as some preliminary developer notes.

You can find a preview of the documentation [here](https://murrayrm-biocrnpyler.readthedocs.io/en/update_docs-12aug2025/).

PR #313, opened 2025-09-14
Automatic versioning and new v1.3.2 ready for release 

ayush9pandey:
After recent merges from PRs #311 and #312 a new release was not set. This PR updates the package version so it can be released. In that process, I automated the versioning for the package and added instructions in `docs/develop.rst` for all future releases. 
This PR will also enable users to run `biocrnpyler.__version__` to check which version they are running.

Additional fixes/changes:
- Drop python support for 3.7 and 3.8 in setuptools in pyproject.toml (as discussed in #298)
- Remove setup.py since it is deprecated now (pyproject.toml is the replacement)
- Update docs/conf.py to get version from _version.py (automatically generated file when running `pip install .`)
- Updated "refactor-modules" from docs/conf.py to "main"


ayush9pandey:
If the changes look good, we can merge and then release v1.3.2

murrayrm (approved):

PR #315, opened 2025-10-14
Code cleanup (ruff)

murrayrm:
This PR reformats all code in the BioCRNpyler package using `ruff` to create a consistent coding style throughout the repository.  To keep things simple, this uses the [Black code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html), which has been optimized for "consistency, generality, readability and reducing git diffs" (from the documentation)

While (almost?) every file in the repostory has changed, there are no actual changes to the code.  The only changes are formatting.

With this change, the following commands can be run to maintain a consistent code style:

* `ruff format`: look through all files and make changes to be compatible with the Black code style.
* `ruff check`: check for consistency with a variety of coding standards, including pycodestyle, pyflakes, isort, and others.

In reviewing this PR, I suggest quickly looking through the changes to see if there are any updates that you think might be more confusing than they are worth.  For the most part the Black code style is fixed, but we could potentially make some changes by tweaking some ruff rules, if desired.

### DIFF
--- a/docs/develop.rst
+++ b/docs/develop.rst
@@ -53,13 +53,24 @@ GitHub repository file and directory layout:
       + \*.ipynb - Jupyter notebooks
 
 
-Naming Conventions
-==================
+Naming Conventions and Code Style
+=================================
 
 Generally speaking, standard Python and NumPy naming conventions are
 used throughout the package.
 
 * Python PEP 8 (code style): https://peps.python.org/pep-0008/
+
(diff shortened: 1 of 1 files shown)


ayush9pandey:
I think that we could add the two commands for future linting management to the Developer Notes page: https://biocrnpyler.readthedocs.io/en/latest/develop.html

(its easy enough to look for the commands but with many linting tools out there it might be good to note that we are using `ruff` and the  Black linting style)

ayush9pandey (approved):

PR #316, opened 2025-11-02
Docstring updates

murrayrm:
This PR provides updated docstrings for components, mechanisms, mixtures, and core classes and functions.  Existing docstrings were converted to NumPy format (with numpydoc conventions) and missing docstrings were added, along with examples.

I used Claude to generate the docstrings and then checked them all to make sure they made sense and that the examples ran.  Here is the prompt that was used for Claude (Opus 4.1 and Sonnet 4.5):

> I would like your help in writing docstrings for a Python package called 'BioCRNpyler" (imported as 'bcp').  I have added files from the Git repository to the project so that you have access to the code and user documention.  I am writing the documentaion in NumPy style, using Sphinx and numpydocs.
> 
> For each file in the package, I will attached the current version of the file, which has been modified from what is in the repository,  I would like you to generate properly formatted numpydoc-style docstrings for all classes, methods, and fucntions in the file, following the NumPy doc format and the guidelines in the BioCRNpyler Developer Notes (available in the project files).  You should then insert them directly into the code (for example using str_replace tool) and return a file with the updated docstrings.
> 
> I am attaching two example files (component.py and chemical_reaction_network.py) that are examples from previous sessions.  Please use these as a guide.
> 
> Some additional notes:
> * I use a line length limit of 78 characters.  Please respect this when writing the docstrings.
> * I use single quotes for strings.
> * When you generate numbered or unnumbered lists, make sure to leave a blank line prior to the start of the list, otherwise Sphinx doesn't process it properly.
> * When you describe mechanisms (e.g. 'binding') or parameters (e.g. 'kb' and 'ku') put them in single quotes.
> * Assume that BioCRNpyler has been imported and is available with prefix 'bcp'.
> * Assume that NumPy has been imported and is available with prefix 'np'.
> 
> Confirm that you understand the instructions and then I will give you the files one at a time.

In addition, specialized prompts were used for DNA components, mechanisms, and mixtures:

> DNA components: The first set of files I am going to give you are for DNA components.  These files define a set of Component subclasses that are used to define genetic constructions (DNA --> RNA --> protein).  The main component is a DNA assembly, and so I will start with that one, then give you the files for the different elements (promoters, RBS, coding sequence, etc).

> Mechanisms: The set of files I am going to give you are for mechanisms.  Make sure to list all of the parameters that are required for that mechanism in the notes section, enclosed in single quotes. Make sure to list all of the parameters that are required for the mechanism in the notes section, in single quotes (e.g., 'ku' or 'kb').
> 
> I'm attaching two examples of mechanism files, 'binding.py' and 'global_mechanisms.py', that can be used as a rough template for what I am looking for.  Note in particular that examples are included at the class level, but not within internal methods.

> Mixtures: I would now like to process some mixtures.  Here are some specific instructions:
> * In the Notes section, list the default mechanisms included in the mixture. For each mechanism, specify:
>   - The mechanism type (in single quotes, e.g., 'transcription')
>   - The explicit mechanism class name (in backticks, e.g., `Transcription_MM`)
>   - A brief description of what the mechanism does, including the reaction scheme where appropriate
> * Include Examples at the class level that show how to create a mixture using bcp.DNAassembly components
> * In examples, use the components and parameter_file keywords in the constructor to create the complete mixture in one call
> * Create a "mixture" (not "extract") in the examples
> * Keep examples simple and focused on the main use case

In addition to the updated docstrings, a few changes were made to the code:
* The `degredase` keyword in the ` Deg_Tagged_Degradation` global mechanism was changed to `degradase` (corrected spelling)
* The `AnitDilutionConstiutiveCreation` global mechanism was renamed `AntiDilutionConstitutiveCreation` (corrected spellings)

Some requests:
* @dr3y It would be great if you could do a quick check on the many classes you contributed around enumeration and DNA constructs and make sure the doc strings make sense.
* @zjuradoq It would be great if you could do a quick check on the membrane-related components and mechanisms and make sure they make sense.
* @WilliamIX, @ayush9pandey: Feedback welcome on this choice of format and whether there are other changes we should make to docstrings as part of this PR.

### DIFF
--- a/biocrnpyler/core/chemical_reaction_network.py
+++ b/biocrnpyler/core/chemical_reaction_network.py
@@ -81,6 +81,11 @@ class ChemicalReactionNetwork(object):
     Once created, species and reactions cannot be removed, only added. This
     ensures CRN validity is maintained throughout its lifetime.
 
+    Chemical reaction networks can be simulated by writing the output
+    as SMBL using `write_sbml_file` and then loading into an external
+    simulator, or by using the bioscrape package, which can be called
+    directly using `simulate_with_bioscrape_via_sbml`.
+
     Examples
     --------
     Create a simple CRN manually:


biocrnpyler/components/basic.py
```diff
@@ -281,6 +656,18 @@ def update_species(self) -> List[Species]:
         return species
 
     def update_reactions(self) -> List[Reaction]:
+        """Generate binding and unbinding reactions for the complex.
+
+        Uses the 'binding' mechanism to generate reactions for complex
```
ayush9pandey: Similar to the previous comment, I am not sure if the component "generates a reaction". The code below gets the mechanism to associate it with the component and then the mechanism generates the reactions. I understand that this is perhaps minor but it's also core to how biocrnpyler was designed.

biocrnpyler/core/chemical_reaction_network.py
```diff
@@ -26,20 +26,104 @@
...
-    stochastic propensity =
-    .. math::
-    k \Prod_{inputs i} (S_i)!/(S_i - a_i)!
-    where a_i is the spectrometric coefficient of species i
+    r"""Container for chemical species and their reactions.
+
+    A ChemicalReactionNetwork (CRN) represents a biochemical system as a set
+    of species and reactions between them. CRNs can be compiled to SBML format
+    for simulation with various tools, or simulated directly with bioscrape or
+    roadrunner.
```
ayush9pandey: Do we want to add here that bioscrape simulation is built into biocrnpyler..

biocrnpyler/components/basic.py
murrayrm: Also: do we want to use similar wording for `update_species`?

biocrnpyler/components/basic.py
```diff
@@ -9,114 +9,381 @@
...
 
 
 class RNA(Component):
-    """RNA sequence of a given length.
-
-    Notes:
+    """RNA sequence component with specified length.
+
+    An `RNA` component represents an RNA sequence with a given length in
+    base pairs. This is a basic component that produces no reactions on its
```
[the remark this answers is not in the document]
ayush9pandey: Yes, that is a good re-wording.

biocrnpyler/components/basic.py
ayush9pandey: I like either the first or the second possibilities you listed out: ""Call 'binding' mechanism to generate component reactions."  or "Use...."

and yes, we might similarly re-word `update_species` in the components.

murrayrm:
I've addressed @ayush9pandey's comments.  I've got some more documentation updates coming (for the reference manual), so my plan is to merge this tomorrow (Sun) unless someone (e.g., @dr3y, @zjuradoq, or @WilliamIX) let me know they would like me to wait so that they can look through things (definitely not urgent; these are just docstrings and we can modify them later).

ayush9pandey (approved):
Thank you for all the changes. Looks great!!

PR #322, opened 2026-05-10
Default mechanisms in Components + Update PURE mixture, mechanisms, and parameters

murrayrm:
~DRAFT PR.  Putting this out so people can see the changes coming.  The items that were still outstanding are now done:~

- [x] Need to add an example of `plot_all_species_containing`
- [x] May want to update the way attributes passed as a single string are handed (see below)
- [x] Need to add user documentation for PURE and plotting changes

This PR makes a number changes to the way that PURE is implemented:
* There is now a single `PURE` mixture that uses parameters to turn on/off the use of machinery, resources, and energy (`include_machinery`, `include_resources`, `include_energy`).  This means that we don't need separate `SimplePURE`, `EnergyPURE`, etc.  A separate `fuel` argument names the species that regenerates ATP, and setting it to `None` keeps ATP consumption but drops the regeneration pathway.  The old `BasicPURE` mixture is now deprecated: it is `PURE` with all three switches on and `fuel=None`, which is the energy model it had before `PURE` was introduced, and it issues a `DeprecationWarning` when constructed.
* Parameter values have been updated to come from documented sources (mainly SK07 = V. Sotiropoulos and Y. N. Kaznessis.  "Synthetic Tetracycline-Inducible Regulatory Networks: Computer-Aided Design of Dynamic Phenotypes.", BMC Systems Biology 1, no. 1 (2007))
* The unused `ndps` argument has been removed from `PURE`.  It was accepted and documented but never created a species, so `PURE(ndps=...)` silently did nothing; it now raises a `TypeError`.

In addition, some changes that go beyond PURE were also implemented:
* There are now default mechanisms for all standard components.  This means that you can create models without having to explicitly specific the mechanisms, as long as you are OK with the defaults.
* New plotting routines to generate plots containing a given species (`plot_all_species_containing`) or gene expression data (`plot_gene_expression_data`).  How much of each species name appears in the legend is controlled with `show_material`, for the species itself and the totals, and `show_complex_material`, for the complexes containing it.
* Updated use of `attributes` to always pass a list (rather than a string).  This addresses the issue pointed out by @zjuradoq in PR #320.  In addition, an attribute passed as a single string is now treated as one attribute, rather than being split into one attribute per character.
* User documentation for the PURE mixture and the plotting routines has been added to the User Guide, and the plotting routines now appear in the reference manual.
* `Mixture.compile_crn` no longer calls `resetwarnings()`.  It was discarding every warning filter in the process, not just ones set by the package, so filters installed by the caller were dropped the first time a CRN was compiled.  This affects all mixtures, not just PURE.
* Unit tests for `PURE` have been added in `Tests/Unit/test_pure.py`, covering the three switches, the species name arguments, and the `BasicPURE` deprecation.

Sample plot for `plot_gene_expression_data` (from `gfp_expression.py`):
image

Sample plot for `plot_all_species_containing` (from `gfp_expression.py`):
image

### DIFF
--- a/biocrnpyler/__init__.py
+++ b/biocrnpyler/__init__.py
@@ -11,6 +11,10 @@ from .mixtures import *
 
 # All utilities
 from .utils import *
+from .utils.plotting import plot_all_containing as plot_all_containing
+from .utils.plotting import (
+    plot_gene_expression_data as plot_gene_expression_data,
+)
 
 try:
     from ._version import version as __version__
--- a/biocrnpyler/__init__.py
+++ b/biocrnpyler/__init__.py
@@ -11,11 +11,9 @@ from .mixtures import *
 
 # All utilities
 from .utils import *
-from .utils.plotting import (
-    plot_all_species_containing as plot_all_species_containing,
-)
-from .utils.plotting import (
-    plot_gene_expression_data as plot_gene_expression_data,
+from .utils.plotting import (  # noqa: F401
+    plot_all_species_containing,
+    plot_gene_expression_data,
 )
 
 try:
--- a/biocrnpyler/__init__.py
+++ b/biocrnpyler/__init__.py
@@ -11,7 +11,9 @@ from .mixtures import *
 
 # All utilities
 from .utils import *
-from .utils.plotting import plot_all_containing as plot_all_containing
+from .utils.plotting import (
+    plot_all_species_containing as plot_all_species_containing,
+)
 from .utils.plotting import (
     plot_gene_expression_data as plot_gene_expression_data,
 )
--- a/biocrnpyler/mechanisms/txtl_parameters.tsv
+++ b/biocrnpyler/mechanisms/txtl_parameters.tsv
@@ -19,7 +19,7 @@ translation		ktl	0.05	/sec	15 aa/s and protein length of 300 aa
 #
 # PositiveHillTranscription and NegativeHillTranscription
 #
-# We computing the various rates here based on the the following paper:
+# We compute the various rates here based on the the following paper:
 #
 #   [SK07] V. Sotiropoulos and Y. N. Kaznessis.  "Synthetic
 #   Tetracycline-Inducible Regulatory Networks: Computer-Aided Design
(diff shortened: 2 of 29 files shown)


biocrnpyler/mechanisms/txtl_parameters.tsv
```diff
@@ -16,31 +16,45 @@ gene_expression		kexpress	19.23	/sec
 transcription		ktx	0.05	/sec	50 nt/s and transcript length of 1000 nt
 translation		ktl	0.05	/sec	15 aa/s and protein length of 300 aa
 
+#
 # PositiveHillTranscription and NegativeHillTranscription
-transcription		K	1.6e-3	uM	RMM (TODO)
-transcription		k	3.25	/sec/uM	Tx_cat from Singhal (= ktx?? TODO)
+#
+# We computing the various rates here based on the the following paper:
```
ayush9pandey: Typo: "We are computing"

biocrnpyler/__init__.py
```diff
@@ -11,6 +11,12 @@
 
 # All utilities
 from .utils import *
+from .utils.plotting import (
+    plot_all_species_containing as plot_all_species_containing,
+)
+from .utils.plotting import (
+    plot_gene_expression_data as plot_gene_expression_data,
+)
```
ayush9pandey, line 17: Wouldn't this simpler import line do the same?
`from .utils.plotting import plot_gene_expression_data`

(same for the other one)..

examples/gfp_expression.py
ayush9pandey, line 1: Since all other examples are jupyter notebooks, do you think we should transform this example also into a notebook?

ayush9pandey:
Looks good to me..I added a few comments above.

examples/gfp_expression.py
murrayrm: I put this in as a plain Python file intentionally, since it was intended as more of developer example (helping me make sure parameters made sense) rather than a user example.  I can see the argument both ways and we might want to rethink how we handle examples.

There was also a PR where @ayush9pandey converted this and added things to it, so if we convert to a notebook we should add those things back in.

I've created issue #339 to have a discussion about this.

biocrnpyler/__init__.py
murrayrm: This was due to ruff: if you import a specific function and don't use that in a file, you get an [F401 error](https://docs.astral.sh/ruff/rules/unused-import).  I fixed it in a different way that is less ugly.

PR #334, opened 2026-08-09
BioCRNpyler support for numpy v2.0+

ayush9pandey:
BioCRNpyler build restricts numpy<2, since Bioscrape required this in the past for running simulations. Bioscrape is now compatible with numpy>2, so this PR proposes to remove the numpy<2 requirement for biocrnpyler.

Will wait for all tests to pass and then merge.


murrayrm (approved):
Looks good.

PR #342, opened 2026-08-26
Fix GFP loading comparison by updating mRNA degradation rates + related bug fixes

murrayrm:
This PR fixes a problem that was showing up in `examples/gfp_expression.py` in which the expression level for GFP was going _up_ when competing CFP DNA was added, rather than down (due to resource loading).  In chasing down the bug (with help from Claude Opus), it turns out that the RNase parameters were incorrect and the result was that adding CFP DNA soaked up some of the excess RNAse and so GFP expression rose (so there was loading, just not on the expected resources).

This PR fixes that problem by updating the mRNA degradation parameters to more realistic values (from the literature).

The GFP/CFP fix exposed a series of additional bugs in the way that default parameter files were being included in various mechanisms, so those are included here as well (possibly better as a separate PR, but everything is linked together, so proposing this as one).

## Fix 1: Set mRNA degradation parameters from measured TX-TL lifetimes

In `examples/gfp_expression.py`, figure 4 adds a competing gene (CFP) to show
resource competition. It showed the opposite: adding CFP *raised* per-gene GFP (dashed = GFP only, solid = GFP + CFP):

Figure_4_old

The blue curves (on top of each other) are as expected, since the `SimpleTxTlExtract` doesn't include any resource limits.  The green curves (`EnergyTxTlExtract`) also work: energy usage is higher with CFP, so we get less GFP.  But the orange curves (`TxTlExtract`) didn't make sense: GFP expression is _higher_ (by 2X!) with CFP DNA added.

The cause was in the RNase parameters. With `kb = 1 /sec/uM`, `ku = 1.25E-6 /sec` and `kdeg = 1.3E-4 /sec`, the Michaelis constant is

    K_M = (ku + kdeg) / kb = 0.13 nM

against an mRNA pool of a few nM, so RNase was 15-150x saturated and behaved as a near-stoichiometric sink rather than a catalyst: only 6% of it was free with a single gene. Adding a second gene starved degradation further, extending every transcript's lifetime, so per-gene GFP expression went _up_.

Fixing this involved a sequence of changes:

### Degradation rates are now pinned to measurements

The values for `kdeg` and `K_M` are now set using `k_eff`, which is the effective rate of mRNA degradation:

| | value | source |
|---|---|---|
| extract | `k_eff = 8.25E-4 /sec` (20.2 min mean lifetime) | MN19, determined by assay (Fig. S4) |
| cell | `k_eff = 0.002 /sec` (8.3 min) | SK07 Table 1, R44/R90 |

(SK07 = Sotiropoulos and Kaznessis, 2007.) TX-TL 2.0 measures 17.45 min for the same transcript MN19 measures at 20.2 min, so two independent measurements bracket a narrow range.

Only `k_eff` is fixed by data. Depending on the mechanism, other values also need to be determined.

For the Michaelis-Menten model,`K_M = 8 uM` is carried over from MN19 as a modeling convention — they state the constants "were chosen so as to obtain kdeg,m determined by the assay" — and `kdeg` follows from `k_eff`, `K_M` and `[RNase]`. This reproduces MN19's V_max exactly: `2.64E-2 * 0.25 uM = 6.60 nM/s` against their stated `6.6 nM/s`.

For the explicit RNase mechanism, we have to include binding and unbinding.  `kb` is not separately identifiable: binding equilibrates in well under a second against an mRNA lifetime of tens of minutes, so only the ratio matters. Holding `K_M` fixed, any `kb` from 0.01 to 100 /sec/uM gives the same lifetime and the same steady-state protein to within 0.1%. It is now documented as the generic protein association rate (BioNumbers 112534) rather than left uncited.

### Fix ribosome concentration in extract: 10 uM -> 1.1 uM

MN19 best-fits `Rtot = 1100 nM` for active ribosomes in a TXTL reaction, and estimates 1450-2500 nM total. The cytoplasm-to-reaction dilution is not a single factor, so each machinery species now records its own basis rather than a uniform 1/10 rule. RNAP stays at 0.5 uM, which is MN19's stated maximum and coincidentally also 1/10 of the cell value.

### Ribosome-bound mRNA is now degradable

`TxTlExtract` and `EnergyTxTlExtract` passed `recursive_species_filtering=False` to `Degradation_mRNA_MM`, which exempts mRNA inside the Ribo:mRNA complex from degradation entirely. Since mRNA spends most of its life in that complex, RNase could not reach it. Enabling recursive filtering adds one species and two reactions, with no nested RNase:RNase regress, and makes the mRNA lifetime insensitive to the ribosome dissociation constant rather than dependent on it. Interesting, the docstrings had already indicated the bound mRNA was also degraded, but the mixtures did not actually support that.

### Translation constants

The unbinding rate of ribosomes to mRNA for `translation_mm` and `energy_translation_mm` had been set to values from SK07, but they were using the wrong constants.  In particular, SK07 R46 was being used as the reference.  This reaction is

    Rib:mRNA --> Rib:mRNA1 + mRNA      100 /sec

which is the ribosome clearing the binding site to form an elongating complex and releasing the mRNA for the next ribosome (polysome loading), _not_ dissociation. SK07 has no unproductive-dissociation step, so it supplies no `ku` for `Translation_MM`. These values were overridden by the mixture parameter files, so this is not a behavior change — but reading R46 as `ku` weakens the complex ~270,000x, and anyone "fixing" the shadowing (e.g, specifying a new parameter file that didn't set 'ku') would have collapsed protein yield.

RNAP binding in the various parameter files is now harmonized to SK07 R31/R32 (`kb = 8.6`, `ku = 0.01`) across extract and PURE, replacing `ku = 2.5E-06` marked "source unclear (TODO)". R31/R32 *are* a genuine binding/unbinding pair. RNAP is in large excess over DNA in every mixture, so `TxTlExtract` moves 0.8%, everything else under 0.001%.

## Fix 2: Four additional issues found along the way

In chasing this down, a number of other issues came to light:

**Updated `kexpress`.** For one step gene expression,`kexpress` is derived (`ktx * ktl / RNA loss rate`) and is declared in three files. These needed to be updated to reflect the the new degradation rates. Mixtures override it, so nothing misbehaved.

**`cell_parameters.tsv` had no `rna_degradation` entry.** `SimpleTxTlDilutionMixture` degrades RNA with a `Dilution` mechanism *named* `rna_degradation`, but that instance carries `mechanism_type = 'dilution'` whatever it is named, so the `(mechanism_type, part_id, param)` tier matched the growth dilution rate. RNA was removed twice at 5.78E-4 /sec instead of degradation plus dilution. That is a well-formed hit rather than a fallback, so no warning was raised. `kexpress` in the same file already assumed the correct total of 2.578E-3, so the file was internally inconsistent by 2.2x.

**`Transcription_MM` and `Translation_MM` dropped `parameter_file`.** Both accept `parameter_file='mechanisms/txtl_parameters.tsv'` but never passed it to `MichaelisMentenCopy.__init__`, so the parent fell back to `enzyme_parameters.tsv` and the txtl defaults were never read. Those catalysis defaults are unreachable from these classes anyway — lookups go through `mechanism=self`, so only `transcription_mm`/`transcription` keys can match. A scan confirmed these were the only two mechanisms in the codebase with this bug; the other 20 all forward correctly.

**Cell machinery was declared with `part_id='e_coli'`.** Initial concentrations are looked up with `part_id` set to the mixture's own name, so those entries only resolved for a mixture literally named `e_coli`. The cell mixtures default to `name=''`, so the machinery stayed at zero and no protein was expressed, with no warning — falling through to zero is the normal path for an undeclared species. `e_coli` appears nowhere in the Python source.  This was fixed by removing `e_coli` from the parameter file (so it matches generically unless a parameter with a more specific `part_id` is provided.

The last two compounded: `TxTlDilutionMixture` orginally could not compile at all, and once it could, it expressed nothing. It has been unusable for some time, and nothing in the test suite noticed.

## Fix 3: Tests to stop these issues from recurring

Unit tests were added to exposre the bugs above.  In addition `Tests/Unit/test_parameter_consistency.py` adds four checks:

* duplicated parameters must agree unless the key is allowlisted with a reason (nine keys, covering degradation rates and machinery levels that genuinely differ between cells, extract and PURE);
* a mechanism default must match a value some mixture actually uses — the allowlist alone permits mixtures to disagree, which would otherwise hide a stale default, and this is what the `kexpress` bug was;
* a mixture using a mechanism whose parameter is allowlisted must set that parameter in its own file;
* every mixture compiles to a CRN with a protein species and non-zero machinery, since a missing parameter can leave a species at zero rather than raising.

Run against the commits that introduced the fixes, they fail at `771cd33` (four failures), `245cce4` (two), `aa53305` and `ae96fba` (one each), and pass at `b71a8a0`. Every defect above is caught at the commit before its fix.

## Fix 5: `gfp_expression.py` extract comparisons

Coming back to the starting point for this entire branch, the point of Figure 4 in `gfp_expression.py` was to show the differences between the different mixtures.  Originally GFP and CFP were both at 1 nM concentration, which wasn't enough to distinguish between `SimpleTxTl` and 'TxTlExtract` since there were no loading effects.

By raised the DNA concentration to 10 nM per gene, where the ribosome pool actually becomes limiting, it was possible to separate out the expression levels from the different mixtures:

| | alone | w/ CFP | |
|---|---|---|---|
| simple | 618 | 618 uM | 1.00x — models no machinery, load invisible |
| regular | 575 | 482 uM | 0.84x — ribosomes shared |
| energy | 25 | 13 uM | 0.50x — fixed metabolite pool split evenly |

Two corrections were also needed for the comparison to mean anything: the dashed "alone" traces reused results simulated at 1 nM, and the PURE section shares the `cfp_initial_conditions` variable. Figures 1-3 stay at 1 nM, since they compare mixtures at a realistic plasmid concentration while figure 4 deliberately pushes into the resource-limited regime. The vertical axis is truncated at 200 uM so the energy traces stay readable, with final values quoted in the legend.

Here's what all of that did in terms of the plots that started this whole adventure (note flip in line style, which I thought looked better):
Figure_4_new

## Fix 6: PURE documentation

This also seemed like a good time to confirm that there were no similar issues in the PURE mixture(s).

PURE keeps no RNA degradation mechanism, and the reasoning is now recorded rather than left implicit. PURE is reconstituted from purified components and carries no ribonucleases. The detailed PURE reaction network of Jurado, Pandey and Murray (2023) has no mRNA degradation step either, and their MGapt measurements show transcript still accumulating past two hours; Jurado and Murray (2024) attribute loss of MGapt signal in commercial PURE to the dye chemistry rather than to mRNA decay. No quantitative mRNA half-life for PURE appears in the literature, so no rate is invented.

The consequence — mRNA accumulating and progressively binding ribosomes — is intended: once ATP is exhausted the ribosomes stall on transcript they can no longer translate. This is now noted in the PURE docstring.

PURE's concentrations are deliberately unchanged. RNAP at 1.3 uM, Ribo at 3 uM and the metabolite levels come from the Nucleus and OnePot PURE protocols, which are separate sources from the TX-TL extract measurements.

## Open issues

* One pre-existing `ODEintWarning` from the energy mixture at 10 nM. Checked against an independent tight LSODA solve: 12.74443 uM against 12.74443 uM, so it is cosmetic, and it is left in place.
* Pre-existing in `examples/gfp_expression.py` on main: an unused `sec` import (ruff F401) and a `ruff format` divergence in the PURE plotting section. `ruff format` wants to change none of the lines added here.
* 23 parameters remain duplicated across files with identical values. They are in sync and the new tests will catch it if they drift, but the underlying redundancy is unchanged.

## AI assistance disclosure

This PR was drafted with Claude Code (Claude Opus 5) working interactively with @murrayrm.

The AI was used to: locate the saturated-RNase root cause and the four subsequent defects; retrieve and convert the literature values (Garamella 2016, Marshall & Noireaux 2019, Sotiropoulos & Kaznessis 2007, Jurado/Pandey/Murray 2023, Jurado/Murray 2024, BioNumbers 108604/111532/111927/112534); run the parameter sweeps, identifiability, convergence and regression checks quoted above; and draft the parameter-file comments, the new tests, the commit messages and this description.

All parameter values, the choice of which measured lifetime to standardise on, and the modeling decisions were directed and reviewed by @murrayrm. Every numerical claim above was produced by a simulation or an arithmetic check rather than asserted.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,3 +69,21 @@ they were used, and what is AI generated.
 
 You may draft the text of issues and pull requests. The contributor reviews and
 edits that text before it is submitted.
+
+### Copyright headers
+
+New source files start with:
+
+```
+#  Copyright (c) <current year>, BioCRNpyler Developers. All rights reserved.
+#  See LICENSE file in the project root directory for details.
+```
+
+Use the year the file is created, and the collective name rather than an
+individual or institution - the same convention NumPy and SciPy use
+("NumPy Developers", "SciPy Developers"), which avoids having to revisit
+headers as contributors change.
+
+Leave the headers on existing files alone. Many still name Build-A-Cell and
+carry the year they were written, and rewriting them is churn that buries the
+real change in a diff.
--- a/Tests/Unit/test_parameter_consistency.py
+++ b/Tests/Unit/test_parameter_consistency.py
@@ -1,0 +1,194 @@
+#  Copyright (c) 2026, BioCRNpyler Developers. All rights reserved.
+#  See LICENSE file in the project root directory for details.
+
+"""Consistency checks across the default parameter files.
+
+The same parameter is often declared in several files: a mechanism default
+in `mechanisms/*_parameters.tsv` and again in each `mixtures/*_parameters.tsv`
+that wants a different value.  Nothing keeps those copies in step, so a value
+updated in one file can silently go stale in another.  The tests here pin down
+which divergences are deliberate and check that everything else agrees.
+"""
+
+import glob
+import io
+import os
+from collections import defaultdict
+
+import pytest
+
+import biocrnpyler as bcp
+
+# Parameters that are deliberately different between files, with the reason.
+# Anything else that differs is treated as drift and fails the first test.
+INTENTIONAL_DIVERGENCES = {
+    # Cells lose RNA to growth dilution as well as degradation, so the
+    # one step expression rate is lower than in a cell free extract.
+    ('gene_expression', '', 'kexpress'),
+    # Degradation is faster in vivo than in extract, so both the
+    # Michaelis-Menten constants and the first order rate differ.
+    ('rna_degradation', '', 'kdil'),
+    ('rna_degradation_mm', '', 'kdeg'),
+    ('rna_degradation_mm', '', 'ku'),
+    # Machinery and metabolite levels are properties of the system being
+    # modelled: whole cells, TX-TL extract, and reconstituted PURE each
+    # have their own measurements.
+    ('initial concentration', '', 'ATP'),
+    ('initial concentration', '', 'NTPs'),
+    ('initial concentration', '', 'RNAP'),
+    ('initial concentration', '', 'RNase'),
+    ('initial concentration', '', 'Ribo'),
+}
+
+# Files holding mechanism defaults rather than the settings of one mixture.
+MECHANISM_FILES = {
+    'binding_parameters.tsv',
+    'enzyme_parameters.tsv',
+    'transport_parameters.tsv',
+    'txtl_parameters.tsv',
+}
+
+# Mixtures are exercised with a single constitutively expressed gene.
+MIXTURES = [
+    'ExpressionExtract',
+    'SimpleTxTlExtract',
+    'TxTlExtract',
+    'EnergyTxTlExtract',
+    'PURE',
+    'ExpressionDilutionMixture',
+    'SimpleTxTlDilutionMixture',
+    'TxTlDilutionMixture',
+]
+
+
+def parameter_files():
+    root = os.path.dirname(
+        os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    )
+    pattern = os.path.join(root, 'biocrnpyler', '**', '*.tsv')
+    return sorted(glob.glob(pattern, recursive=True))
+
+
+def parameters_by_key():
+    """Map (mechanism, part_id, param_name) to {file name: value}."""
+    table = defaultdict(dict)
+    for path in parameter_files():
+        for line in io.open(path, encoding='utf-8'):
+            if line.startswith('#') or not line.strip():
+                continue
+            fields = line.rstrip('\n').split('\t')
+            if len(fields) < 4 or fields[0] in ('mechanism', 'mechanism_id'):
+                continue
+            key = (fields[0], fields[1], fields[2])
+            table[key][os.path.basename(path)] = fields[3]
+    return table
+
+
+def make_gene():
+    return bcp.DNAassembly(
+        name='gfp', promoter='pconst', rbs='rbs_strong', protein='GFP'
+    )
+
+
+def concentration(entry):
+    """Initial concentrations come back as Parameters or as plain numbers."""
+    return float(getattr(entry, 'value', entry))
+
+
+def test_parameter_files_are_readable():
+    # Guards the file based tests, which quietly pass if nothing is parsed
+    table = parameters_by_key()
+    assert len(parameter_files()) >= 5
+    assert len(table) > 50
+
+
+def test_no_unintended_divergence():
+    """Duplicated parameters agree unless the difference is deliberate."""
+    drifted = {}
+    for key, values in parameters_by_key().items():
+        if (
+            len(set(values.values())) > 1
+            and key not in INTENTIONAL_DIVERGENCES
+        ):
+            drifted[key] = values
+    assert not drifted, (
+        "These parameters are declared in several files with different "
+        "values. Either bring them back into agreement, or add the key to "
+        f"INTENTIONAL_DIVERGENCES with a reason: {drifted}"
+    )
+
+
+def test_mechanism_defaults_are_not_stale():
+    """A mechanism default must be a value some mixture actually uses.
+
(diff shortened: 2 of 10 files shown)


ayush9pandey (approved):
Since this PR changes many parameters in the database, I think we should re-run all example and tutorial notebooks so that the notebook outputs that are displayed on Github use these updated parameters. We could do it here in this PR or another one that follows this.

Issue #338, opened 2026-08-14
Allow lists of multiple substrates for Membrane Components (i.e. Transport)


zjurado:
I am building a model using compartment/membrane components, speciifally  `MembraneChannel`, that can facilitate the transport of multiple different substrates. 

Currently, it appears that the component only accepts a single substrate. To model a single channel that transports multiple different species (like a general porin or a non-specific transporter), I have to instantiate a separate channel component for each substrate, then compile them independently and add reactions to create a complete model.  

It would be highly useful if membrane/transport components could accept a list of species for the substrate argument.

**Example of desired behavior:**
```python
import biocrnpyler as bcp
# ... (setup environment and species)
s1 = bcp.Species("glucose")
s2 = bcp.Species("arabinose")
channel_protein = Species("alphaHL")

# Current limitation (requires separate components, manual compilation, and added reactions):
# channel_1 = bcp.MembraneChannel(channel=channel_protein, substrate=s1)
# channel_2 = bcp.MembraneChannel(channel=channel_protein, substrate=s2)

# Proposed feature:
multi_channel = bcp.MembraneChannel(channel=channel_protein, substrates=[s1, s2])

Issue #333, opened 2026-07-30
Use consistent naming conventions for classes


murrayrm:
The current code uses a lot of different naming conventions for classes.  For example, just in the area of mechanisms, we have:
* mechanisms/binding.py:class Two_Step_Cooperative_Binding
* mechanisms/enzyme.py:class MichaelisMentenReversible
* mechanisms/global_mechanisms.py:class Degradation_mRNA_MM
* mechanisms/global_mechanisms.py:class Deg_Tagged_Degradation
* mechanisms/txtl.py:class Transcription_MM
* mechanisms/txtl.py:class Translation_MM
* mechanisms/txtl.py:class multi_tx
Camel case (Python), lower case with underscores (C), mixtures of the two, abbreviations (MM, tx, Tx), etc.

We should decide on a more consistent convention for classes.  Let's use this issue to sort out what makes sense and then we can document a consistent set of guidelines and update all of the class names (with legacy names still allowed for backward compatability).

murrayrm:
For mechanisms, one possibility would be a pattern that looks like this: MechType_primary_secondary.  So, for example:
* Binding_TwoStep_Cooperative: a type of 'binding' mechanism
* Catalysis_MichaelisMenten: a type of 'catalysis' mechanism
* Catalysis_MichaelisMenten_Reversible: a type of 'catalysis' mechanism
This doesn't exactly follow Python conventions (which would be pure camel case), but the first word would match the mechanism name, the next word(s) would describe the level of detail of that mechanism, and any additional words (optional) would provide additional functionality.

murrayrm:
I asked Claude to go through all 35 mechanism classes and trace how the names are used at runtime. Summary of what it found:

### There are three name spaces here, not one

Every mechanism carries three independently-chosen names, and today all three drift apart:

| | example | who consumes it |
|---|---|---|
| **class name** | `Transcription_MM` | users writing Python |
| **`mechanism_type`** | `'transcription'` | dict key for `mechanisms={...}` lookup |
| **instance `.name`** | `'transcription_mm'` | **column 1 of the parameter TSVs** |

[...]

PR #343, opened 2026-09-11
Refactor of Membrane Transport Components and Mechanisms

zjuradoq:
## Description
This PR introduces a comprehensive refactor of the membrane transport architecture, improving component input flexibility, standardizing mechanism nomenclature, and adding robust support for active and facilitated transport modeling.

### Component Updates
* **Global Component Update**: Expanded input parsing across *all* membrane components (including `MembraneSensor`, `MembranePump`, `MembraneChannel`, etc.) to safely accept `str`, `Species`, and `Component` objects as inputs, seamlessly extracting and formatting species regardless of the input type.
* **`DiffusibleMolecule`**: Allows the definition of multiple permeable molecules during a single instantiation rather than requiring individual calls. Natively accepts `Species` inputs and enforces compartment assignment based strictly on the component's defined compartments (safely overriding the default compartment of the provided species).
* **`IntegralMembraneProtein`**: Removed the `direction` attribute. Mechanism type changed to `membrane_integration` (from `membrane_insertion`). Remains the primary tool for instantiating generic membrane proteins prior to defining specific subclasses.
* **`MembraneChannel`**: 
  * Changed mechanism type to `diffusion`.
  * Removed the `direction` attribute (simple channels operate bidirectionally).
  * Removed legacy substrate/product differentiation logic (diffusion mechanisms no longer distinguish between substrates and products beyond their compartment).
  * Retained list-based user definitions for size-based segregation rather than forcing a global `size` attribute onto all small molecules.
* **`MembraneCarrier` (New)**: Introduced to accurately model uniporters, antiporters, and symporters. Dynamically utilizes either `diffusion` or `transport` mechanisms depending on the specific carrier being modeled.
* **`MembranePump`**: Updated to strictly require a `direction` attribute (`importer` or `exporter`) to automatically designate the correct internal and external compartments for substrate movement. 
* **`MembraneSensor`**: Expanded input parsing to safely accept `str`, `Species`, and `Component` objects.

### Mechanism Updates & Nomenclature Standardization
All mechanisms have been renamed to follow a strict `MechType_primary_secondary` naming convention.

* **`Diffusion_Simple`** (formerly `Simple Diffusion`)
* **`Diffusion_Facilitated_Channel`** (formerly `Simple Transport`)
* **`Integration_MembraneProtein`** (formerly `Membrane Protein Integration`)
* **`Diffusion_Facilitated_Carrier`** (formerly `Facilitated Transport MM`): 
  * Replaced the Heaviside function with a steep Sigmoid (Logistic) curve to prevent negative concentrations and simulation crashes. 
  * Redesigned reaction architecture following Deng 2014 (GLUT1 model) to include 4 distinct configurations (`in` and `out`). Dropped the generic `product` designation in favor of relative substrate compartmentalization.
* **`Transport_SecondaryActive_Symporter` (New)**: Requires a dictionary of driving ions (`{'mol': 'ratio X:sub'}`). Allows multiple ions or isotopes to drive the same substrate. Features fully reversible mass-action reactions, allowing users to reverse transport direction by adjusting kinetic rates.
* **`Transport_SecondaryActive_Antiporter` (New)**: Models counter-transport. Requires a strict string-based dictionary separated by `:` to dictate driving ion and substrate stoichiometry.
* **`Transport_PrimaryActive_ABCexporter`** (formerly `Primary Active Transport MM`): Upgraded to accept `Species`, `Component`, and `str` inputs. Reaction steps updated to include reversible transitions and refined unbinding kinetics for substrates and waste (ADP).
* **`Sensor_TwoComponentSystem`** (formerly `Membrane Signaling Pathway MM`): Added the dimerization step for activated NarL, which is now properly output as the final species in the pathway.

### Testing & Documentation
* Updated all mechanism and component unit tests to reflect the new naming conventions and input typing.
* Refactored Examples 7 and 8 to demonstrate the implementation of the new carrier, pump, and channel architectures.
* Example 9 to demonstrate transport of molecules across different mechanims architectures.

### Schematics
* Schematics for Examples 7 and 8, contained in one folder

(Formatted and copy edited using Gemini based on provided notes)


zjuradoq:
This PR is now Ready for Review. It addresses Issue #338  and Issue #333 (regarding diffusion and transport mechanisms).

**Note on Test Failures:**
I noticed some checks were unsuccessful. Based on my understanding, the failure appears to be caused by an architecture mismatch in the compiled C-extension for libsbml.

I would appreciate any guidance or recommendations on how to resolve this in the workflow configuration.

ayush9pandey:
It seems that python 3.9 support is dropped now in python-libsbml. I created an issue on their repository https://github.com/sbmlteam/python-libsbml/issues/45 to ask the question. If they have dropped it, which seems to be the case because a wheel for that version was not distributed in the latest release, then we would also need to update our CI to run for >= 3.10.

zjuradoq:
Thank you @ayush9pandey for reaching out to the libsmbl repo. I see from the issue, they have dropped python 3.9 support.

ayush9pandey:
I've pushed a branch that removes python v3.9: https://github.com/BuildACell/bioCRNpyler/pull/345
Once the tests pass on that PR, we should merge it and then let the tests run on this PR again.

biocrnpyler/components/membrane.py
```diff
@@ -103,24 +96,48 @@ def __init__(
...
+            else:
+                name = f"MultiDiffusible_{internal_compartment.name}"
 
         Component.__init__(
-            self=self,
-            name=name,
-            attributes=attributes,
-            default_mechanism=Simple_Diffusion(),
-            **kwargs,
+            self=self, name=name, attributes=attributes, **kwargs
```
ayush9pandey, line 140: Just for me to understand -- does this component not need an initialization of default_mechanism to `Diffusion_Simple`. I understand that a mechanism could be specified at the mixture level and it will work correctly.

zjuradoq:
> I've pushed a branch that removes python v3.9: #345 Once the tests pass on that PR, we should merge it and then let the tests run on this PR again.

I have approved and merged #345.

biocrnpyler/components/membrane.py
zjuradoq: I have not been in the habit of providing a default mechanism but I think that providing one at the component level as a default should be required. I will update all membrane components.

PR #345, opened 2026-09-23
Drop python v3.9 support in BioCRNpyler CI

ayush9pandey:
`python-libsbml` stopped supporting python v3.9 (https://github.com/sbmlteam/python-libsbml/issues/45) since it is EOL. So, we'll need to remove v3.9 from our CI so that it runs. I have left
```
requires-python = ">=3.9"
```
in pyproject.toml as it is for now.


zjuradoq (approved):
python-version 3.9 has been successfully removed from the CI.

